PR #2, opened 2015-09-20
Added the deal.II-cdr code as a gallery example.

drwells:
I think I included everything you asked for (it was a good exercise in CMake). This probably is not an ideal first submission because of its size (four programs and a library) but I think people will be interested in the `cmake` structure.


bangerth:
I've been struggling with the format of these directories. Each of your four programs are not quite individually free-standing (though they have individual `author`, `dependencies`, etc. files) but they're not separate either. After thinking about this a bit, may I convince you to move the four directories up by one level, call them (for example) `cdr-1` through `crd-4` (or call them differently)? Clearly, the `common` directory needs to go somewhere, for which I would suggest it go into `cdr1/common` and the other directories then simply refer to the files there via `../cdr-1/common`.

What I want to achieve is a way to process individual programs individually in the scripts that build documentation out of each of these programs. I could alternatively just consider all of your files as a single project that ends up on one very large page. If that's what you prefer, then please simply move the `author` and `dependencies` files up one level to the top-level dir of your project.

drwells:
I do not mind moving things around (I will have to rewrite the CMake code though).

Let me make sure I understand what you are proposing. Should the directory structure (starting at the root of the repository) be
```
/cdr-1/
/cdr-1/common/
/cdr-2/
/cdr-3/
/cdr-4/
```

instead of
```
/1/
/1/common/
/2/
/3/
/4/
```
?

bangerth:
I'll leave the naming of the directories to you, but I think it's ok to use more descriptive names. You can also use `/convection-diffusion-reaction-1/` if you prefer, or any other name.

drwells:
I am still working on this, but redoing the directory structure is far harder than I thought it would be since the five projects are no longer controlled by one `CMakeLists.txt` file. I don't know when I will be finished, but I am still trying to get this sorted out.

drwells:
@bangerth I thought about this some more, and the only way to satisfy the layout we agreed on is to package components separately. This gets weird quickly; a user would have to compile multiple gallery units to run one example. One of my goals was to show how to create one package with multiple files, so I think it would be best to only include one of the solvers and the library code behind it in the gallery.

bangerth:
OK. I would say please do as you think is most useful. If you think it makes the most sense, we can also go back to your original directory layout. It may come out slightly awkward when I generate the doxygen pages, but I think we can explain things as necessary in the accompanying text.

drwells:
I think this is the best approach because it matches the (proposed) format of the other gallery examples. I think the optimal solution is to keep the common format here and point interested readers at the other solvers if they are interested; I'll update the README with a link.

bangerth:
That seems sensible. Let me know when you're there.

drwells:
I think my last commit (I `fixup`ed it in) should finish things; as far as I can tell this is done.

bangerth:
OK, thanks. I've merged this and we'll just see what happens when I try to run my scripts on it. I might come back asking for changes if I find that there is something that would be easier if changed.

PR #7, opened 2015-12-22
A classical approach to solving the cook membrane problem

jppelteret:
In this work we take a classical approach to solving the equations
governing quasi-static finite-strain compressible elasticity, with code
based on `step-44`. The formulation adopted here is that seen in many
texts on solid mechanics and can be used as the starting point for
extension into many topics such as material anisotropy, rate dependence
or plasticity, or even as a component of multi-physics problems.


Quasi_static_Finite_strain_Compressible_Elasticity/README.md
```diff
@@ -0,0 +1,117 @@
...
+finite element formulations for solid mechanics. It is typically used to 
+test for and demonstrate the shear-locking (or locking-free) behaviour of 
+a finite element ansatz under quasi-incompressible conditions. 
+As it is so widely referred to in the literature on finite-strain elasticity, 
+we reproduce the example here.
+However, we consider on the compressible case to avoid many of the complexities
+that arise in `step-44`, which provides an efficient approach to deal with
+the quasi-incompressible case.
+
+A classical approach to solving the cook membrane problem.
```
davydden: perhaps make it a header, e.g. `####`

.gitignore
```diff
@@ -26,3 +26,4 @@
 *.exe
 *.out
 *.app
+/.project
```
davydden: this is Eclipse-related, as far as i recall. I would keep this file unchanged to restrict changes in this PR to the elasticity-related code.

Quasi_static_Finite_strain_Compressible_Elasticity/README.md
```diff
@@ -0,0 +1,117 @@
...
+In this work we take a classical approach to solving the equations governing 
+quasi-static finite-strain compressible elasticity, with code based on 
+`step-44`. The formulation adopted here is that seen in many texts on solid 
+mechanics and can be used as the starting point for extension into many 
+topics such as material anisotropy, rate dependence or plasticity, or even as
+a component of multi-physics problems.
+
+The basic problem configuration is summarised in the following image.
+A beam of specific dimensions is fixed at one end and a uniform traction load
+is applied at the other end such that the total force acting this surface
```
davydden: i suppose
> acting ON this surface

Quasi_static_Finite_strain_Compressible_Elasticity/README.md
```diff
@@ -0,0 +1,117 @@
...
+* Version 8.2.1 or greater of `deal.II`
+
+
+## Compiling and running
+Similar to the example programs, run
+```
+cmake -DDEAL_II_DIR=/path/to/deal.II .
+```
+in this directory to configure the problem.  
+You can switch between debug and lease mode by calling either
```
davydden: release

Quasi_static_Finite_strain_Compressible_Elasticity/README.md
```diff
@@ -0,0 +1,117 @@
...
+well documented and therefore does not need to be reproduced here. 
+We refer the reader to, among many other possible  texts, Holzapfel (2001) and 
+Wriggers (2008) for a detailed description of the approach applied in this work. 
+It amounts to a reduction and slight reworking of `step-44` (accounting for 
+the removal of the two additional fields used therein). We also refer the reader 
+to `step-44` for a brief overview of the continuum mechanics and kinematics 
+related to solid mechanics.
+
+## Results
+The following results were produced using the following material properties:
```
davydden: maybe change one of the `following` to a synonym?

Quasi_static_Finite_strain_Compressible_Elasticity/README.md
```diff
@@ -0,0 +1,117 @@
...
+the removal of the two additional fields used therein). We also refer the reader 
+to `step-44` for a brief overview of the continuum mechanics and kinematics 
+related to solid mechanics.
+
+## Results
+The following results were produced using the following material properties: 
+* Shear modulus is 422.5kPa 
+* Poisson ratio is 0.3
+
+The discretised reference geometry is looks as follows:
```
davydden: > is looks as

davydden:
Thanks a lot, JP. 

I have very minor suggestions. Could not look at the code as GitHub does not show it, but I am sure it's all good!

jppelteret:
@davydden Thanks a lot for all of your recommendations. If you want to view the source code on Github, you can do it by navigating to the `Files changed` tab, then going to the hidden source ` Quasi_static_Finite_strain_Compressible_Elasticity/cook_membrane.cc
` and then clicking on the view button on the right-hand side of the source pane.

davydden:
@jppelteret i won't be able to comment on it in such a case. Not that i would need it, though :smile: I quickly went through the code, i have no further suggestions. :+1:

Quasi_static_Finite_strain_Compressible_Elasticity/README.md
bangerth: acting ON this surface

Quasi_static_Finite_strain_Compressible_Elasticity/README.md
```diff
@@ -0,0 +1,117 @@
...
+The basic problem configuration is summarised in the following image.
+A beam of specific dimensions is fixed at one end and a uniform traction load
+is applied at the other end such that the total force acting this surface
+totals 1 Newton. Displacement in the third coordinate direction (out of plane)
+is prevented in order to impose plane strain conditions.
+
+![Problem geometry](./doc/problem_setup.png)
+
+Note that we perform a three-dimensional computation as two-dimensional 
+elasticity corresponds to neither plane-strain nor plane-stress conditions.
```
bangerth: I think that's not quite correct -- one just has to choose a particular 2d formulation that represents 3d conditions. It's true that if you write down the "obvious" 2d equivalent of the 3d equations, then you get neither plane strain nor plane stress. But that may be too subtle an argument :-)

Quasi_static_Finite_strain_Compressible_Elasticity/README.md
```diff
@@ -0,0 +1,117 @@
...
+![Problem geometry](./doc/ref_grid.png)
+
+And an example of the displaced solution is given in the next image.
+
+![Displaced solution](./doc/disp_soln.png)
+
+Below we briefly document the tip displacement as predicted for different
+discretisation levels and ansatz for the displacement field.
+A direct comparison of the following results can be made with those found 
+in Miehe (1994).
```
bangerth: Out of curiosity (and it may be worth adding this here): Do the results you show match what others have found and stated in the literature?

bangerth: And does this correspond to the "locked" or "locking-free" case?

bangerth:
My apologies for only now looking at this -- a nice extension, and a very nicely written program!

I added a couple of belated comments for which you should have gotten emails. Here are a couple other things about the program (since I can't inline comments there), not because I want you to change them but because I noticed them while learning something from your program:

* I think something like `StandardTensors` might as well be in deal.II if you'd like to move it there.
* The internal members of `Material_Compressible_Neo_Hook_One_Field` don't need to be `protected` because you're not deriving from this class. (Or is it your intention that someone could?) You can make them `private`.
* In `PointHistory` you store a pointer to such a material. I'd change this to a `std_cxx11::unique_ptr` so you don't have to manage the memory yourself. But beyond this, the question arises why you need a pointer to begin with. I believe it's because you can't initialize the memory already when you create the `PointHistory` object. But why can't you? Don't you already know the global parameters at the point where you create these objects? If so, maybe pass them to the constructor of `PointHistory`, rather than have a `setup_lqp` function?
* `Solid::make_grid` is only going to work in 3d because you call `Point<dim>(x,y,z)`. So the check for `dim==3` before that seems unnecessary. Alternatively, you could use `dim==3 ? Point<dim>(x,y,z) : Point<dim>(x,y)`.
* Instead of `print_conv_header/footer` and doing it all by hand, you could use the `TableHandler` class.
* Nice application of `WorkStream` -- our paper on the subject was just accepted ;-)

Quasi_static_Finite_strain_Compressible_Elasticity/README.md
jppelteret: Fair enough. I've fixed the note to get the point across without going into detail.

jppelteret:
@bangerth Thanks for all of your comments. We really just lazily refactored `step-44` to make an easier starting point to finite strain elasticity, but thanks for your compliments :-) I've now addressed most of the issues (I think). Regarding some of the other points you've raised:
* The way that we construct and initialise the `PointHistory` class is a mixture of (a) "for historical reasons" (i.e. how its done in `step-18` so it a familiar approach) and (b) because we want whoever wishes to extend this program not to have to struggle to conceptualise how to incorporate hetrogeneous media and multiple constitutive laws. I agree that it could be done much better (i.e. https://github.com/dealii/dealii/issues/1438 or an enhancement thereof), so perhaps we can come back to this when I start putting together a unified interface for QP data storage and manipulation.
* As is clear in the code, most of the ingredients for the `StandardTensors` already exist in `deal.II`, so are you suggesting that having a static instance of these tensors would be of all-round benefit to most users? Apart from the fact that we call them fairly frequently, we had originally just wanted them to have a similar name to that used in the solid mechanics literature.
* I typically construct my own tables because I personally find it beneficial to print them out in real time rather than once-off at the end of each timestep. Is there some way to get the `TableHandler` to do the same (I don't think so) or is it conceptually possible to extend it to do this?
* Congrats on the paper acceptance. Always a nice way to end the year!

bangerth:
@jppelteret -- thanks.

Re standard tensors: yes, I think putting this into the library would be ok. I had thought about something similar earlier on (but didn't think of using the surrounding class) and decided that having symbols such as `I` or `II` is just not good practice for global function names. But I see no reason not to do it in a contained class.

Re tables: I see your point. I don't think there currently is a way to print incremental updates. It probably wouldn't be terribly hard to implement, but I get that it may be more than you want to do right now.

jppelteret:
@bangerth I've made a note to myself to do both of these. I'm sure that the `StandardTensors` extension won't take long so I'll get onto that sooner rather than later. 

However, as you've indicated, the tables extension won't be a priority for me (although I am interested in doing it). I think that some basic refactoring of `TableHandler::write_text` to split the header and data output and adding a function such as `TableHandler::add_and_write_text` might be most of what is needed. I'll have to look to the tests to see the expected behaviour of each of the formatting types to gauge this better. I expect that there will be basic issues with padding and formatting which will take time to sift through, which is why I can't do this immediately.

bangerth:
I was rather thinking of something of the kind `TableHandler::write_text_last_line_only()` (or maybe adding a boolean flag to the function signature that indicates whether the whole table, or only its last line, will be written). The cheapest way to implement it would be to just call `write_text()` for the whole table, put the result into an `ostringstream` and extract the last line of output.

jppelteret:
Ok, well this would be easy to do so I’ll gladly implement that. 

Maybe I’m crazy in how I do things, but for debugging purposes I use the output from each stage of the assembly and solve process for each newton iteration as a real-time status update. This is what I would like to do for myself, and would involve printing the table entries as soon as they’re stored. Would you be happy to accept such a change as well as a full last-line output?

bangerth:
Ah, I see. Maybe printing every entry as its entered is too fine a distinction. Let's not worry about it then :-)

PR #8, opened 2015-12-28
Added parameter file; code cleanup; fixed small bug in post-processing

jppelteret:
I forgot to do the following to the cook membrane problem before the final commit:

* Add the parameter file
* Clean up some small dregs that remained from debugging
* Fix a minor issue regarding the construction of a custom quadrature rule with which to extract the solution at a given point.

All of these issues have been attended to here. Apologies :-/


davydden:
no limit to perfection :+1:

PR #14, opened 2016-01-09
Update cdr

drwells:
I spent a few hours going over this entry and it now looks a lot nicer when rendered with the new scripts.


cdr/common/include/deal.II-cdr/parameters.h
```diff
@@ -1,11 +1,16 @@
...
 #include <deal.II/base/parameter_handler.h>
 
+#include <string>
+
+// I prefer to use the ParameterHandler class in a slightly different way than
+// usual: The class Parameters creates, uses, and then destroys a
+// ParameterHandler inside the <code>read_parameter_file</code> method instead
+// of keeping it around. This is nice because now all of the run time
+// parameters are contained in a simple class and it can be copied or passed
+// around very easily.
```
bangerth: That's not an unreasonable approach. We use it in `ASPECT` as well.

cdr/common/include/deal.II-cdr/system_rhs.templates.h
```diff
@@ -17,12 +26,12 @@ namespace CDR
   void create_system_rhs
   (const DoFHandler<dim>                                 &dof_handler,
    const QGauss<dim>                                     &quad,
-   const std::function<Tensor<1, dim>(const Point<dim>)> convection_function,
-   const std::function<double(double, const Point<dim>)> forcing_function,
+   const std::function<Tensor<1, dim>(const Point<dim>)>  convection_function,
+   const std::function<double(double, const Point<dim>)>  forcing_function,
```
bangerth: Out of curiosity, why do you not pass it by reference?

cdr/solver/cdr.cc
```diff
@@ -102,27 +104,27 @@ CDRProblem<dim>::CDRProblem(const CDR::Parameters &parameters) :
   n_mpi_processes {Utilities::MPI::n_mpi_processes(mpi_communicator)},
   this_mpi_process {Utilities::MPI::this_mpi_process(mpi_communicator)},
   fe(parameters.fe_order),
-  quad(3*(2 + parameters.fe_order)/2),
+  quad(parameters.fe_order + 2),
```
bangerth: :-)

cdr/solver/cdr.cc
```diff
@@ -102,27 +104,27 @@ CDRProblem<dim>::CDRProblem(const CDR::Parameters &parameters) :
...
-          pcout (std::cout,
-                 (Utilities::MPI::this_mpi_process(mpi_communicator)
-                  == 0))
+    [](const Point<dim> p) -> Tensor<1, dim>
+      {Tensor<1, dim> v; v[0] = -p[1]; v[1] = p[0]; return v;}
+  },
+  forcing_function
+  {
+    [](double t, const Point<dim> p) -> double
+      {
```
bangerth: Nice use of lambdas!

cdr/common/include/deal.II-cdr/system_rhs.templates.h
drwells: I remember reading a stack overflow post where the author said one should pass these by value, but I cannot find that post now. It works fine (and makes more sense) when passed by reference, so that advice (if it ever existed) was probably taken out of context. I will fix this.

cdr/solver/cdr.cc
drwells: I did not understand how the quadrature orders worked until I rewatched that lecture.

PR #15, opened 2016-01-11
Pass std::function instances by reference.

drwells:
Thanks @bangerth for the comments on the last PR, including pointing out the strange choice of passing these by value.


bangerth:
Glad to help. I don't think there was anything wrong with your previous choice, but since `std::function` objects can be big, it seemed prudent not to copy them more often than necessary.

PR #16, opened 2016-02-14
Add printable names to each entry.

drwells:
In response to dealii/dealii#2180:

Presently there are three code gallery entries and each one has a slightly different naming convention. This commit adds a 'printable' name to each entry which will make the presentation uniform.

@jppelteret @sfloresm I did my best to guess these names from your contributions. Let me know (or open a PR) if you would like something else.


jppelteret:
Perfect. Thanks very much @drwells!

PR #20, opened 2016-06-12
CeresFE ready to push

elstargazer:


CeresFE/CMakeLists.txt
```diff
@@ -0,0 +1,50 @@
+##
+#  CMake script for the step-22 tutorial program:
```
bangerth: please update this comment with the name of your program, or if it is a small variation of the one from step-22, say that it is a variation.

CeresFE/doc/authors
```diff
@@ -0,0 +1,2 @@
+Roger Fu <rogerfu@fas.harvard.edu>
+Anton Ermakov <eai@mit.edu>
```
bangerth: I will admit that it is pretty dumb, but it confuses cmake that there are no newlines at the end of the file for all of the files in `doc/`. Would you be willing to add those?

PR #21, opened 2016-09-08
Multiphase Navier-Stokes solver

manuel-quezada:
Hi, 

I want to contribute to the code-gallery with a small project about two phase flow simulations. 

Thanks, 

Manuel Quezada


two_phase_flow/CMakeLists.txt
```diff
@@ -0,0 +1,57 @@
+##
+#  CMake script for the step-40 tutorial program:
+##
+
+# Set the name of the project and target:
+#SET(TARGET "TestLevelSet")
+#SET(TARGET "TestNavierStokes")
+SET(TARGET "MultiPhase")
```
bangerth: Can you update these comments and remove things you don't need any more?

bangerth:
Well, this isn't a *small* project :-)

Two questions:
* What's in the zip file? Is it necessary?
* If you have pictures or movies, then we'd be happy to take those as well. I think they may be good advertisements for what the program actually does.

two_phase_flow/CMakeLists.txt
manuel-quezada: I will modify this file. The Naver-Stokes and the Level Set solvers are completely independent from each other. I decided to include code to test each of those solvers by themselves for few scenarios. I figured some people might be interested in just one or the other.  That is the reason why I have these lines 

#SET(TARGET "TestLevelSet")
#SET(TARGET "TestNavierStokes")

in the cmake file. Then people can just uncomment accordingly depending on what they want to run. I explain this in detail in the readme file.

manuel-quezada:
Hi Professor, here are the answers to your questions: 
- The zip file have some figures. 
- I have prepared more figures and some animations as well. Do you want me to place them all within the same zip file or do you prefer me to create a normal folder (within doc or somewhere else?)?

two_phase_flow/CMakeLists.txt
bangerth: Ah, I see now. Yes, a description in the Readme.md would be great. Also add a comment in front of the current place that references the readme for more information on this.

bangerth: Alternatively -- do these 3 programs share any code, or are they self-contained sitting next to each other in the same directory? If the latter, would it make sense to just have 3 code gallery programs?

bangerth:
Regarding the zip file: just unpack it and add the individual figures somewhere under `doc/`. This way we can link them into the documentation.

As for how many figures you add, I'll leave this up to you. It's mostly a question of file size -- I have few hesitations to add 5 or 10MB, but would like to not add more than that per code gallery program.

two_phase_flow/CMakeLists.txt
manuel-quezada: They share code. There are basically two solvers: NavierStokesSolver and LevelSetSolver. These solvers can't be run by themselves (they have no main) and they are independent from each other. They really just know how to solve their corresponding equation for a single time step and they have no knowledge on geometry, the problem itself, the other solver, etc. 

There are three codes to run different experiments: 
-TestNavierStokes performs 2D and 3D convergence tests on the NS Solver. 
-TestLevelSet performs two simple tests on the LS solver. 
-MultiPhase combine both solvers to solve two phase flow problems.

I had this organization since it was easy not to mess with the solvers once I knew they were working and just concentrate on creating the geometry, boundary conditions and so on for the different tests. Also, if we had some new idea we wanted to test (say for the level set) we could do it by just concentrating on LevelSetSolver and TestLevelSet and not mess with anything else. Finally, during the multi-phase simulations It was reassuring to have the tests for the NS and LS always there since I could just re-run those tests at any point if I had doubts in the correctness of the code. 

I explain all this and more in the readme file. I have changed the CMakeLists, committed and pushed the changes. Please let me know any other change you might want me to do.

manuel-quezada:
Done. I have deleted the zip file and add the folder doc/results where I added figures and animations for three problems. I made sure to be under 10MB. The whole project is ~8.7MB. 

Please let me know any more changes you want me to do. Maybe extend something on the readme file, add more references or whatever you think convenient. Thanks.

bangerth:
I've taken a look at this and am a bit ambiguous about the many pictures. The existing code gallery programs display *all* pictures at the top, see for example https://dealii.org/developer/doxygen/deal.II/code_gallery_Quasi_static_Finite_strain_Compressible_Elasticity.html . This won't work with this many pictures. It's also not quite clear to me what *additional* value each picture brings. What would you think of selecting a few, representative ones? And/or combining them into a movie that we could embed? (I think half a dozen pictures + a movie may be the sweet spot.)

two_phase_flow/Readme.md
```diff
@@ -0,0 +1,92 @@
+# Two Phase Flow #
+-----------------------------------
```
bangerth: This doesn't seem to render correctly via markdown. It shows both of the hash marks, but still marks it up as a header. Do you need the hash marks if you do the minusses in the next line?

manuel-quezada: I eliminated the hash marks and I think it is working now. Let me know otherwise.

manuel-quezada:
I picked few (~5-7) pics for each problem (breaking dam, falling drop and filling a tank) and also left the animations. All this is within doc/results. Maybe we can even remove two of the problems and just concentrate on one. If you want to do this, I will eliminate everything except the results for the breaking dam problem (I like that problem the most). What do you think?

bangerth:
I think it's ok. We're quite ok with taking codes as they are, including multiple variations. 

Can you please squash the commits into one so that the files that are no longer part of the pull request aren't part of the commits we will merge?

manuel-quezada:
Done (or so I think). I did:

_git rebase -i master_

and changed all the commits (except the first one) to squash. I got this:

_Successfully rebased and updated refs/heads/my-code-gallery-project._

But when I do git log I can still see all previous commits. Did I miss something?

bangerth:
You mean your *local* git log? What branch are you now on?

manuel-quezada:
I am at "my-code-gallery-project". If I do git log being on that branch I see the following: 

> commit 8ec7670ed347af51a00f30a01b76b6c897743377
> Author: Manuel Quezada de Luna <mql@Manuels-MacBook-Pro-2.local>
> Date:   Thu Sep 8 16:57:33 2016 -0500
> 
>     My project: two phase flow simulations
>     
>     add zip file with plots of some simulations
>     
>     update con CMakeLists.txt and on animations and figures
>     
>     Minor chance in CmakeLists.txt
>     
>     remove previous results
>     
>     Add fewer images and minor modification in Readme.md
>     
>     Change in Readme.md
>     
>     Incompressible two-phase interaction

The very last line is indeed the message I included during git rebase -i master. Oh, I think I know what is happening. I think I forgot to comment (#) the previous commit messages. I can rebase again to fix this problem. I have another question. After I rebase, do I have to:

_git push origin my-code-gallery-project_ 

again? Or do I have to do something else after rebase? Thanks! :)

bangerth:
No, what you see is the *one commit message* that has been generated by concatenating the commit messages of the commits you squashed. This is all right. (But it would have been ok to throw some of them away and only leave those that really matter -- because, after all, you can no longer take this commit apart and so it no longer matters that you "removed the previous results".)

After all rebase operations, you have to do `git push -f origin my-code-gallery-project`. Note the `-f` because you are overwriting the previous content of the branch at github.

manuel-quezada:
Yes, I forgot to comment the previous messages. I figured after I finished the rebase. Anyway, I was able to modify the message for the single commit by doing git commit --amend and then I forced push as you said. I think all worked out properly. Thanks!

bangerth:
Fantastic. Thanks!

PR #24, opened 2017-02-23
updated ceres.cc to include output of core-mantle boundary

elstargazer:


CeresFE/src/ceres.cc
```diff
@@ -684,6 +685,14 @@ void StokesProblem<dim>::assemble_system() {
 		Assert(
 				local_quadrature_points_history < &quadrature_point_history.back(),
 				ExcInternalError());
+		
+		double cell_area = cell->measure();
+		if(cell_area<0)
+			append_physical_times(-1);
+		AssertThrow(cell_area > 0
```
bangerth: So if you throw an error if `cell_area` is negative, why does the previous statement make sense?

CeresFE/src/ceres.cc
```diff
@@ -726,12 +735,18 @@ void StokesProblem<dim>::assemble_system() {
...
 				double &local_old_phiphi_stress =
 						local_quadrature_points_history[q].old_phiphi_stress;
 				double r_value = fe_values.quadrature_point(q)[0];
 				double z_value = fe_values.quadrature_point(q)[1];
+				
+				// if(system_parameters::present_timestep == system_parameters::initial_elastic_iterations)
+				// {
+				// 	old_stress *= 0;
+				// 	local_old_phiphi_stress = 0;
+				// }
```
bangerth: did you mean to leave this commented out piece of code in here?

CeresFE/src/ceres.cc
elstargazer: Yeah, this is just to write a last line to the "physical_times.txt" output file so that it records that this last step did not terminate properly.  Else our plotting/analysis code that reads the physical_times.txt and other output files sees an incomplete set of output files for that step and gets confused.  I can remove if you prefer, since it doesn't affect how the deal.ii code is run.

CeresFE/src/ceres.cc
elstargazer: No, I thought that was gone. Sorry!

CeresFE/src/ceres.cc
bangerth: OK, no worries. I guess if this was deal.II code, we'd put a comment in front of it that says something like "Put an error marker in the output file if we get a cell with negative area, then abort the code."

bangerth:
Let me know when you have the updated code in the pull request.

elstargazer:
Ok, everything should be in order for merging to code-gallery.

PR #29, opened 2017-03-21
Linear Elastic Active Skeletal Muscle Model

jppelteret:
In this example we present an implementation of a linearised active
skeletal muscle model with application to the simulation concentric
contraction of the human biceps brachii.


bangerth:
Why did we not merge this right away? I haven't read all of it, but it seems like the kind of submission we want to see!

jppelteret:
Its all good :-) I think that I added this during a busy patch as the release was being prepared.

PR #32, opened 2017-11-13
Time-dependent Navier-Stokes

Jie-Cheng:
Time-dependent Navier-Stokes with Grad-Div stabilization and IMEX scheme.


bangerth:
Ah, I look forward to having this program in the code gallery!

@tjhei, @kronbichler  -- do either of you have time to take a look at this program? If not, let me know and I can do it.

time_dependent_navier_stokes/Readme.md
```diff
@@ -0,0 +1,138 @@
...
+        {\tilde{S}}^{-1}B{\tilde{A}}^{-1} & -{\tilde{S}}^{-1} \\
+      \end{array}
+    \right)
+@f}
+where ${\tilde{S}}$ is the Schur complement of ${\tilde{A}}$, which can be decomposed 
+into the Schur complements of the mass matrix, diffusion matrix, and the Grad-Div term:
+@f{eqnarray*}
+    {\tilde{S}}^{-1}
+    \approx 
+    {S_{mass}}^{-1} + {S_{diff}}^{-1} + {S_{Grad-Div}}^{-1}
```
tjhei: You are moving the sum outside the inverse operation here. I don't think this approximation is necessary at least to get the Grad-Div term separated in the next equation.

time_dependent_navier_stokes/Readme.md
```diff
@@ -0,0 +1,138 @@
...
+    \approx 
+    [B^T (diag M)^{-1} B]^{-1} + \Delta{t}(\nu + \gamma)M_p^{-1}
```
tjhei: Are you really using diag(M)? It would be helpful to specify this in the documentation here.

time_dependent_navier_stokes/time_dependent_navier_stokes.cc
```diff
@@ -0,0 +1,1008 @@
...
+    timer.leave_subsection();
+  }
+
+  // @sect4{NavierStokes::setup}
+
+  // A helper function to determine which constrint to use based on the current timestep
+  template <int dim>
+  const ConstraintMatrix &NavierStokes<dim>::get_constraints() const
+  {
+    return time.get_timestep() == 0 ? nonzero_constraints : zero_constraints;
```
tjhei: Is there a reason for solving for the update instead of for the full solution at the current time? If you do that, you don't need two different kind of constraints.

time_dependent_navier_stokes/time_dependent_navier_stokes.cc
```diff
@@ -0,0 +1,1008 @@
...
+  template <int dim>
+  std::pair<unsigned int, double> NavierStokes<dim>::solve_linear_system(bool update_precondition)
+  {
+    const ConstraintMatrix &constraints_used = get_constraints();
+
+    if (update_precondition)
+    {
+      timer.enter_subsection("Precondition linear system");
+
+      preconditioner.reset();
```
tjhei: I think these reset() calls are all redundant.

time_dependent_navier_stokes/time_dependent_navier_stokes.cc
```diff
@@ -0,0 +1,1008 @@
...
+    data_out.write_vtu(output);
+    timer.leave_subsection();
+  }
+
+  // @sect4{NavierStokes::process_solution}
+
+  // This function is used to calculate the drag and lift coefficients on the cylinder.
+  // We first calculate the traction of the fluid, which is nothing but the product of the
+  // stress tensor and the normal of the cylindrical surface, and then integrate it along
+  // the cylindrical surface and negate it.
```
tjhei: It might be worth leaving a note here, that the volume integral method described in John 2004 ("Reference values for drag and lift ...") is more accurate.

bangerth:
I'm afraid I don't have the time this week or next to look at this in more detail :-( 

@Jie-Cheng -- if you don't mind, could you address @tjhei 's comments? I can then try to take another look after Thanksgiving.

time_dependent_navier_stokes/Readme.md
Jie-Cheng: Did you mean that lines 116 and 117 should be removed because the Grad-Div term and diffusion term should not be written as if they are decoupled? But ${\tilde{S}}^{-1} \approx [B (diag M)^{-1} B^T]^{-1} + \Delta{t}(\nu + \gamma)M_p^{-1}$ is correct, right?

time_dependent_navier_stokes/time_dependent_navier_stokes.cc
Jie-Cheng: I solved for the update because I thought non-homogeneous constraint is more computationally expensive/less accurate in dealii because additional Lagrangian multiplier is needed internally. Is it in fact as good as homogeneous constraint? If so I should change this implementation.

time_dependent_navier_stokes/time_dependent_navier_stokes.cc
Jie-Cheng: Fixed.

time_dependent_navier_stokes/time_dependent_navier_stokes.cc
Jie-Cheng: Fixed.

time_dependent_navier_stokes/time_dependent_navier_stokes.cc
tjhei: I am not aware of any cost/accuracy differences.

Jie-Cheng: I changed the code to solve for the full solution at the current time. But it came with one problem: in all the time steps other than the first one, I call `constraints.distribute_local_to_global(local_rhs, local_dof_indices, system_rhs)` to avoid unnecessary assembly of the LHS. Now because I am not solving for the update, this constraints is non-homogeneous, and gives me wrong result. As the documentation put: 

> Note that this function will apply all constraints as if they were homogeneous. For correctly setting inhomogeneous > constraints, use the similar function with a matrix argument or the function with both matrix and vector arguments.

But the point of IMEX is not to assemble LHS all the time. What should I do to in this situation?  Does my words make sense? I can update the PR if that makes my question more clear.

bangerth: Can't you manually iterate over all constraints in the `ConstraintMatrix` and set the elements of the rhs vector that correspond to constrained DoFs to the inhomogeneity of the constraint?

Jie-Cheng: I added the following code after the cell loop:
``` 
if (!assemble_lhs)
  {
     for (unsigned int i = 0; i < dof_handler.n_dofs(); ++i)
        {
            if (constraints.is_constrained())
              {
                 system_rhs(i) = constraints.get_inhomogeneity(i);
              }
        }
   }
```
But the result is still not right. Maybe I did not understand what you meant by
> set the elements of the rhs vector that correspond to constrained DoFs to the inhomogeneity of the constraint

Can you elaborate?

bangerth: Hm, yes, thinking about it, this can't work -- you are dealing with a rhs vector, not a solution vector. So scrap this

Could you output the rhs vector for the two methods and compare them? Are they different only in the rows of nodes on the boundary, or also elsewhere? You may want to look at step-25 and 26, btw, where we deal with a similar problem and call `apply_boundary_values()`. This may not be the right solution for your case, but it may inform you how we do things there.

Jie-Cheng: I did two cases: one case with LHS and RHS assembled and distributed at the same, the other only assembles the LHS at the first time step, and compare the `system_rhs`. It turned out the constrained components in `system_rhs` are 0 in both cases, but those are not constrained were different.  As recommended in the documentation, modifying the matrices and vectors after assembly is not preferred. I would rather to solve for the increment to avoid `apply_boundary_values`.

tjhei: Yes, this can't work as is as the ``distribute_local_to_global`` call will scale the constrained equations with a factor that depends on the local matrix entries, so this is not available if you only build a RHS. I think this means going back to the update formulation is the best approach.

time_dependent_navier_stokes/time_dependent_navier_stokes.cc
bangerth: But they're still here.

bangerth:
@Jie-Cheng -- what is the current status of this? I see that there are still a few open comments for which I'm not sure we've ever gotten to a conclusion. Do you have any open questions, or things you want to work on?

I've added a few language edits to the readme file. If you want to do additional changes, you may want to first pull these from github to your local repository before continuing work.

Jie-Cheng:
@bangerth  I changed my application significantly since I opened this PR. Now I am using fully implicit method instead. The new code is much neater than this. So I am thinking to replace this code with my fully implicit code. Is it OK?

bangerth:
That's your choice -- we are happy to take any code you may have!

bangerth:
@Jie-Cheng -- what's the current status of this program?

Jie-Cheng:
@bangerth I've written a parallel version, and I am tuning it. Will update in this week.

Jie-Cheng:
@bangerth the code has been updated to be parallel. It is part of my bigger project. I am very sorry that I waited for such a long time -- because my code changed so much and I never had time to update my PR. Please review it when you have time. I will address the possible issues ASAP!

MagdaNte:
Hi,
thanks for sharing your code, really helpful!

I'm trying to use that for a case of periodic boundary conditions but when I try to invert the [B^T (diag M)^{-1} B]^{-1} in the Block Preconditioner, it says it's either not invertible or not positive definite. I've tried gmres but it's not working either, any ideas? 

Btw I'm not refining my mesh at all

time_dependent_navier_stokes/Readme.md
MagdaNte: Hi @tjhei and @Jie-Cheng ,

I have a unit square domain and when I use fully periodic boundary conditions I'm not able to get the [B (diag M)^{-1} B^T]^{-1}, it says not invertible. I have increased the number of iterations for the CG solver and I also tried to use gmres, but none of them ways worked.  Any ideas why this could happen or how could I fix it?

Thanks in advance for your help!

Best,
Magda

masterleinad:
@MagdaNte This thread is for discussing the current contribution. You are much more likely to get help with your problem if you are asking in the [deal.II User Group](https://groups.google.com/forum/#!forum/dealii).

bangerth:
@tjhei -- I think it would be nice if we could get this one merged for 9.2. Do you think you might have time to look at it?

tjhei:
sure, I can take a look over the next few days.

tjhei:
I fixed two minor issues, ran the documentation and the code itself and I think it is good to merge. This PR has been laying around way too long. Sorry, @Jie-Cheng and thank you for this contribution.

bangerth:
Yay!

PR #34, opened 2017-12-01
Modernise and add assembly for cook membrane problem using automatic differentiation

jppelteret:
Following a request from a user, using this example I have illustrated how one can use automatic differentiation via Sacado to assemble the linear system. Since the problem is variational one can do this using two approaches: (i) the traditional linearisation of the residual, and (ii) computation of both the residual and its linearisation from the cell contribution to the total potential energy functional. This required a significant change to the way that one interacts with material data at the quadrature point, compute quantities (energies, stresses) that track the sensitivities with respect to cell degrees-of-freedom that are being considered during assembly.

At the same time, I chose to update the code to use some more modern features of deal.II.


jppelteret:
@bangerth ping

bangerth:
I must have been completely confused by the odd spelling of that first word in the subject line of this PR :-)

Getting on it asap.

Quasi_static_Finite_strain_Compressible_Elasticity/parameters.prm
```diff
@@ -1,5 +1,10 @@
 # Listing of Parameters
 # ---------------------
+subsection Assembly method
+  # The number of levels of automatic differentiation to be used in the assembly of the linear system
+  set Automatic differentiation order = 0
```
bangerth: I don't know what you mean by "number of levels of AD". Can you elaborate on this here (or in the declaration of the parameter, and then copy it here)?

Quasi_static_Finite_strain_Compressible_Elasticity/cook_membrane.cc
```diff
@@ -62,6 +63,18 @@
 #include <deal.II/numerics/data_out.h>
 #include <deal.II/numerics/vector_tools.h>
 
+#include <deal.II/base/config.h>
+#if DEAL_II_VERSION_MAJOR >= 9 && defined(DEAL_II_WITH_TRILINOS)
+  #include <deal.II/differentiation/ad.h>
+  #define DEAL_II_SUPPORTS_SACADO_FORMULATION
```
bangerth: You don't own the `DEAL_II_` namespace here. Just call it something different.

bangerth: (Or does this `#define` have an effect on the files included further down below? If so, maybe add a line of comment.)

Quasi_static_Finite_strain_Compressible_Elasticity/cook_membrane.cc
```diff
@@ -79,6 +92,43 @@ namespace Cook_Membrane
...
+    };
+
+
+    void AssemblyMethod::declare_parameters(ParameterHandler &prm)
+    {
+      prm.enter_subsection("Assembly method");
+      {
+        prm.declare_entry("Automatic differentiation order", "0",
+                          Patterns::Selection("0|1|2"),
+                          "The number of levels of automatic differentiation to be used in the assembly of the linear system.");
```
bangerth: see my comment on the .prm file

Quasi_static_Finite_strain_Compressible_Elasticity/cook_membrane.cc
```diff
@@ -79,6 +92,43 @@ namespace Cook_Membrane
...
+      parse_parameters(ParameterHandler &prm);
+    };
+
+
+    void AssemblyMethod::declare_parameters(ParameterHandler &prm)
+    {
+      prm.enter_subsection("Assembly method");
+      {
+        prm.declare_entry("Automatic differentiation order", "0",
+                          Patterns::Selection("0|1|2"),
```
bangerth: Why not use `Patterns::Integer(0,2)`?

Quasi_static_Finite_strain_Compressible_Elasticity/cook_membrane.cc
```diff
@@ -79,6 +92,43 @@ namespace Cook_Membrane
...
+                          "The number of levels of automatic differentiation to be used in the assembly of the linear system.");
+      }
+      prm.leave_subsection();
+    }
+
+    void AssemblyMethod::parse_parameters(ParameterHandler &prm)
+    {
+      prm.enter_subsection("Assembly method");
+      {
+        automatic_differentiation_order = Utilities::string_to_int(prm.get("Automatic differentiation order"));
```
bangerth: `prm.get_int(...)` should do this for you.

Quasi_static_Finite_strain_Compressible_Elasticity/cook_membrane.cc
```diff
@@ -181,11 +231,11 @@ namespace Cook_Membrane
     {
       prm.enter_subsection("Material properties");
       {
-        prm.declare_entry("Poisson's ratio", "0.4999",
+        prm.declare_entry("Poisson's ratio", "0.3",
```
bangerth: I suspect these changes to the material parameters are intentional?

jppelteret: Yes, these values were inherited from the parent tutorial and so the defaults changed in the parameter file.

Quasi_static_Finite_strain_Compressible_Elasticity/parameters.prm
jppelteret: Do you find my modification to this text sufficiently descriptive?

jppelteret:
Thanks a lot @bangerth for giving this a look!

Quasi_static_Finite_strain_Compressible_Elasticity/parameters.prm
```diff
@@ -1,5 +1,13 @@
 # Listing of Parameters
 # ---------------------
+subsection Assembly method
+  # The automatic differentiation order to be used in the assembly of the linear system
+  # Order = 0: Both the residual and linearisation are computed manually.
+  # Order = 1: The residual is computed manually but the linearisation is performed using AD.
+  # Order = 2: Both the residual and linearisation are computed using AD. 
+  set Automatic differentiation order = 0
```
bangerth: This looks good. Please copy the description also into the text you have in the `declare_parameter(...)` call.

PR #35, opened 2018-01-23
Multipoint Flux Mixed Finite Element method with deal.II

eldarkh:
This project is a documented implementation of a higher order multipoint flux mixed finite element methods (MFMFE) for Darcy problem that illustrates the use of the `FE_RT_Bubbles` elements. This implementation includes the local elimination of velocity degrees of freedom for an arbitrary order method in both 2 and 3 dimensions.


MultipointFluxMixedFiniteElementMethods/README.md
```diff
@@ -0,0 +1,99 @@
...
+| Nodal assembly             |     1.14s |       8.8% |
+| Output results             |    0.253s |         2% |
+| Pressure CG solve          |     2.39s |        18% |
+| Pressure matrix assembly   |     1.08s |       8.3% |
+| Velocity solution recovery |   0.0741s |      0.57% |
+| Total time                 |       13s |       100% |
+So one can see that the method solves the problem with 262k unknowns in about 6 seconds, with the rest of the time spent for the post-processing. These results were obtained with 8-core Ryzen 1700 CPU.
+# References
+- <a href="https://arxiv.org/abs/1710.06742"> Higher order multipoint flux mixed finite 
+element methods on quadrilaterals and hexahedra</a>
```
bangerth: Can you add author names and years to each of these entries?

eldarkh:
Great! I would be happy to hear comments and/or changes suggestions, as I assume there might be better ways of implementing this method, or some of its internal functions.

`.Rhistory` is there by mistake (I used RStudio for markdown), I removed it.

MultipointFluxMixedFiniteElementMethods/README.md
```diff
@@ -0,0 +1,99 @@
...
+
+ - The first involves localization of degrees of freedom around each vertex in the grid. Only this type occurs in the lowest order case $k=1$. The number of DOFs that are coupled around a vertex equals the number of facets $n_v$ that share the vertex.  
+ - The second type of coupling is around nodes located on facets, but not at vertices. In 2d, these are edge DOFs. In 3d, there are two cases to consider for this type of coupling. One case is for nodes located on faces, but not on edges. The second case in 3d is for nodes located on edges, but not at vertices.
+ - The third type of coupling involves nodes interior to the elements, in which case only the $dim$ DOFs associated with the node are coupled.
```
bangerth: write this as `\`dim\`` (with backticks, instead of dollar signs).

MultipointFluxMixedFiniteElementMethods/README.md
```diff
@@ -0,0 +1,99 @@
...
+ - The third type of coupling involves nodes interior to the elements, in which case only the $dim$ DOFs associated with the node are coupled. 
+
+Due to the localization of DOF interactions described above, the velocity mass matrix obtained from the bilinear form $(K^{-1} \mathbf{u}_h,\mathbf{v})$, is block-diagonal with blocks associated with the Gauss-Lobatto quadrature points. In particular, in 2d, there are $n_v \times n_v$ blocks at vertices ($n_v$ is the number of neighboring edges), $3 \times 3$ blocks at edge points, and $2 \times 2$ blocks at interior points. In 3d, there are $n_v \times n_v$ blocks at vertices ($n_v$ is the number of neighboring faces), $2n_e \times 2n_e$ blocks
+at edge points ($n_e$ is the number of neighboring elements), $5 \times 5$ blocks at face points, and $3 \times 3$ blocks at interior points.
+
+## Elimination procedure
+The local elimination procedure is done as follows (it is very similar to the Schur complement approach, except everything is done locally). Having a system of equations corresponding to a particular node
```
bangerth: Do you mean to say "node $i$"?

MultipointFluxMixedFiniteElementMethods/README.md
```diff
@@ -0,0 +1,99 @@
...
+@f}
+Here, $A_i$ are small local matrices (full matrices), that are cheap to invert. We also store their inverses as they are further used in velocity solution recovery. With this, the second equation in the system above yields
+@f{align}
+B_i^TA_i^{-1}B_i p = g_i - B_i^TA_i^{-1}f,\nonumber
+@f} 
+where $B_i^TA_i^{-1}B_i$ is a local node's contribution to the global pressure system.
+By following the above steps, one gets the global cell-centered SPD pressure matrix with a compact stencil. After solving for the pressure variable, we use the expression for local velocities above in order to recover the global velocity solution.
+
+## Convergence properties
+While the proposed schemes can be defined and are well posed on general quadrilateral or hexahedra, for the convergence analysis we need to impose a restriction on the element geometry. This is due to the reduced approximation properties of the MFE spaces on arbitrary shaped quadrilaterals or hexahedra that our new family of elements inherits as well. However, on $h^2$-parallelograms in 2d and regular $h^2$-parallelepipeds in 3d, there is no reduction in accuracy.
```
bangerth: -> arbitrarily

bangerth: Can you explain what "h^2-parallelgram" means?

MultipointFluxMixedFiniteElementMethods/mfmfe.cc
```diff
@@ -0,0 +1,1039 @@
...
+// As always the program is in the namespace of its own with
+// the deal.II classes and functions imported into it
+namespace MFMFE
+{
+  using namespace dealii;
+
+  // @sect3{Definition of multipoint flux assembly data structures}
+
+  // The main idea of the MFMFE method is to perform local elimination
+  // of the velocity variables in order to obtain the resulitng
```
bangerth: -> resulting

MultipointFluxMixedFiniteElementMethods/mfmfe.cc
```diff
@@ -0,0 +1,1039 @@
...
+  // mass matrices $A_i$ and the corresponding to them $B_i$.
+  namespace DataStructures
+  {
+    // This will be achieved by assembling cell-wise, but instead of placing
+    // the terms into a global system matrix, they will populate node-associated
+    // full matrices. For this, a data structure with fast lookup is crucial, hence
+    // the hash table, with the keys as Point<dim>
+    template <int dim>
+    struct hash_points{
+      size_t operator()(const Point<dim> &p) const{
```
bangerth: you could as well make this function a `static` function. but then it's going to be the only one in the `struct`, so why not just make it a global, non-member function?

MultipointFluxMixedFiniteElementMethods/mfmfe.cc
```diff
@@ -0,0 +1,1039 @@
...
+        size_t h1,h2,h3;
+        h1 = std::hash<double>()(p[0]);
+        h2 = std::hash<double>()(p[1]);
+
+        switch (dim) {
+          case 2:
+            return (h1 ^ h2);
```
bangerth: You're almost making this function dimension-independent, but you're not covering the 1d case. I know you don't need that, but it would be easy to just add this case as well by moving everything into the switch statement.

MultipointFluxMixedFiniteElementMethods/mfmfe.cc
```diff
@@ -0,0 +1,1039 @@
...
+          default:
+            Assert(false, ExcNotImplemented());
+        }
+      }
+    };
+
+    // Here, the actual hash-tables are defined. We use the C++ STL <code>unordered_map</code>,
+    // with the hash function specified above. For convenience these are aliased as follows
+    template <int dim>
+    using MapPointMatrix = std::unordered_map<Point<dim>, std::map<std::pair<types::global_dof_index,types::global_dof_index>, double>, hash_points<dim>>;
```
bangerth: I think these typedefs would be easier to read if they were called `PointToMatrixMap` or `Point2MatrixMap` since that matches how we speak about them.

MultipointFluxMixedFiniteElementMethods/mfmfe.cc
```diff
@@ -0,0 +1,1039 @@
...
+    const unsigned int n_vel = dim*pow(degree+1,dim);
+    std::unordered_map<unsigned int, std::unordered_map<unsigned int, double>> div_map;
+
+    // One, we need to be able to assemble the communication between velocity and
+    // pressure variables and put it on the right place in our final, local version
+    // of the B matrix. This is a little messy, as such communication is not in fact
+    // local, so we do it in two steps. First, we compute all relevant LHS and RHS
+    for (unsigned int q=0; q<n_q_points; ++q)
+    {
+      Point<dim> p = scratch_data.fe_values.quadrature_point(q);
```
bangerth: this variable and a few further down could probably be made `const`.

MultipointFluxMixedFiniteElementMethods/mfmfe.cc
```diff
@@ -0,0 +1,1039 @@
...
+
+      for (unsigned int i=0; i<dofs_per_cell; ++i)
+        for (unsigned int j=i; j<dofs_per_cell; ++j)
+        {
+          double mass_term = scratch_data.phi_u[i]
+                             * scratch_data.k_inverse_values[q]
+                             * scratch_data.phi_u[j]
+                             * scratch_data.fe_values.JxW(q);
+
+          if (fabs(mass_term) > 1.e-12)
```
bangerth: maybe use `std::abs()`

bangerth: same elsewhere

MultipointFluxMixedFiniteElementMethods/mfmfe.cc
```diff
@@ -0,0 +1,1039 @@
...
+    scratch_data.velocity_matrix.reinit(n_edges,n_edges);
+    copy_data.pressure_matrix.reinit(n_edges,n_cells);
+
+    copy_data.velocity_rhs.reinit(n_edges);
+    scratch_data.pressure_rhs.reinit(n_cells);
+
+    std::set<types::global_dof_index>::iterator vi_it, vj_it, p_it;
+    unsigned int i, j;
+
+    for(vi_it = velocity_indices.at((*n_it).first).begin(), i = 0;
```
bangerth: move the initialization of `i` into the declaration to make this easier to read.

MultipointFluxMixedFiniteElementMethods/mfmfe.cc
```diff
@@ -0,0 +1,1039 @@
...
+    unsigned int n_edges = velocity_indices.at((*n_it).first).size();
+    unsigned int n_cells = pressure_indices.at((*n_it).first).size();
+
+    scratch_data.velocity_matrix.reinit(n_edges,n_edges);
+    copy_data.pressure_matrix.reinit(n_edges,n_cells);
+
+    copy_data.velocity_rhs.reinit(n_edges);
+    scratch_data.pressure_rhs.reinit(n_cells);
+
+    std::set<types::global_dof_index>::iterator vi_it, vj_it, p_it;
```
bangerth: these are counting indices and might be profitably declared as part of the loop in which they are used

MultipointFluxMixedFiniteElementMethods/mfmfe.cc
```diff
@@ -0,0 +1,1039 @@
...
+    unsigned int i, j;
+
+    for(vi_it = velocity_indices.at((*n_it).first).begin(), i = 0;
+        vi_it != velocity_indices.at((*n_it).first).end();
+        ++vi_it, ++i)
+    {
+      for(vj_it = velocity_indices.at((*n_it).first).begin(), j = 0;
```
bangerth: move the declaration of `j` into this block to avoid leaking information outside the range where the variable really is used

MultipointFluxMixedFiniteElementMethods/mfmfe.cc
```diff
@@ -0,0 +1,1039 @@
...
+  void MultipointMixedDarcyProblem<dim>::
+  copy_node_to_system(const DataStructures::NodeEliminationCopyData<dim> &copy_data)
+  {
+    A_inverse[copy_data.p] = copy_data.Ainverse;
+    pressure_matrix[copy_data.p] = copy_data.pressure_matrix;
+    velocity_rhs[copy_data.p] = copy_data.velocity_rhs;
+
+    std::set<types::global_dof_index>::iterator pi_it, pj_it;
+    unsigned int i, j;
+    for (pi_it = pressure_indices[copy_data.p].begin(), i = 0;
```
bangerth: same here -- try to minimize the range in which variables live by declaring them in the smallest possible scope. If a variable starts a new life in another part of the function (i.e., if no information is carried from one part of the function to the other), then declare a separate variable there (possibly with the same name, but semantically different).

MultipointFluxMixedFiniteElementMethods/mfmfe.cc
```diff
@@ -0,0 +1,1039 @@
...
+  copy_node_velocity_to_global(const DataStructures::NodeEliminationCopyData<dim> &copy_data)
+  {
+    std::set<types::global_dof_index>::iterator vi_it;
+    unsigned int i;
+
+    for (vi_it = velocity_indices[copy_data.p].begin(), i = 0;
+         vi_it != velocity_indices[copy_data.p].end();
+         ++vi_it, ++i)
+      vel_solution(*vi_it) += copy_data.vertex_vel_solution(i);
+
```
bangerth: no need for the empty line :-)

MultipointFluxMixedFiniteElementMethods/mfmfe.cc
eldarkh: Regarding this comment and the one above. How do I declare two counting indices of different types in a loop? I think that was the reason I put these outside. I could probably enclose the declaration of the indices and the loops into braces, so they have smaller local scope though.

MultipointFluxMixedFiniteElementMethods/mfmfe.cc
eldarkh: I did not make it a function because I wanted to use typedefs as below, hence I need to pass this hash function object as template parameter for `unordered_map`. As a function it will also work but I would need to pass it as an argument to `unordered_map`'s constructor every time I want to create such a map.
I also thought that `operator` functions cannot be static in C++, I don't think it will compile (I tried and it does not, but I might be missing another reason for it).

MultipointFluxMixedFiniteElementMethods/mfmfe.cc
bangerth: Yes, you can't declare more than one loop index in the `for` header unless they have the same type. (There are ways to do this in C++17, but that's not useful here.)

I enclose these things in braces as appropriate to make the scope small.

MultipointFluxMixedFiniteElementMethods/mfmfe.cc
bangerth: Ah, I see the issue of having to pass the address of the function to the object every time you want to create a map. I did not think of this. Your solution then makes sense.

eldarkh:
I made all the changes as suggested and squashed commits. 
Also, the table with wall-time measurements has changed, this is due to my rerunning the test case with `release` configuration rather than `debug` (I added a remark about it in the test case description in `readme.md`).

bangerth:
Let's merge this then! If others should have comments, then they can always chime in at a later time!

PR #37, opened 2018-03-10
Update code gallery code to make it compile cleanly.

bangerth:
Specifically, remove inclusion of a header file that is (i) no longer around, and
(ii) apparently not needed. Also avoid calling a deprecated function.


masterleinad:
Looks good!

PR #38, opened 2018-03-11
Update the way we use DataPostprocessor in the two-phase-flow code gallery.

bangerth:


bangerth:
In reference to https://github.com/dealii/dealii/issues/6030

masterleinad:
This is pretty straightforward.

PR #39, opened 2018-03-13
Fix compiler errors and warnings

masterleinad:
This fixes all the compiler error and warnings for me compiling with 64-bit indices and a recent developer version. Fixes dealii/dealii#6030.

Admittedly, the indentation looks weird since some of the programs use tabs for indenting. Maybe we should just run our indenting script also for the code gallery.

### DIFF
--- a/CeresFE/src/ceres.cc
+++ b/CeresFE/src/ceres.cc
@@ -185,7 +185,7 @@ private:
 	BlockVector<double> solution;
 	BlockVector<double> system_rhs;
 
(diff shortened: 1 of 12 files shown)


bangerth:
Thanks for doing all this. 

Indeed, the indentation looks a bit wonky. We've usually left it to the authors of these programs to decide what style they want to use, but it indeed makes it a bit more difficult to fix up things like your patch. I don't know how much we want to step on authors' turf, though -- we are pretty explicit in saying that these codes are still provided by the authors, not us, and so I feel awkward if we "claim" these programs by imposing our own style...

PR #41, opened 2018-03-20
Do not make a variable 'static' in the elastoplastic code gallery.

bangerth:
Static variables are only initialized once at the beginning of the run.
Here, the value of a variable depends on a function argument, and so
it will only be computed based on the argument passed to the function
*the first time around*, and will never be updated again in later calls.
This is almost certainly not the desired behavior.


tamiko:
@bangerth @davydden Update?

drwells:
Both versions of this function are going to generate the same code but the patch lets us delete a keyword. Since no one has felt strongly enough about this to comment over the last year lets just merge.

PR #50, opened 2019-01-22
Update to AD component of quasi-static Finite-strain compressible elasticity

jppelteret:
This modernises the AD data structures used in this code-gallery example.

### DIFF
--- a/Quasi_static_Finite_strain_Compressible_Elasticity/cook_membrane.cc
+++ b/Quasi_static_Finite_strain_Compressible_Elasticity/cook_membrane.cc
@@ -1826,65 +1826,107 @@ namespace Cook_Membrane
                                               ScratchData_ASM &scratch,
                                               PerTaskData_ASM &data)
     {
-      // Aliases for data referenced from the Solid class
+      const FEValuesExtractors::Vector &u_fe = data.solid->u_fe;
       const unsigned int &n_q_points = data.solid->n_q_points;
       const unsigned int &dofs_per_cell = data.solid->dofs_per_cell;
       const FESystem<dim> &fe = data.solid->fe;
       const unsigned int &u_dof = data.solid->u_dof;
-      const FEValuesExtractors::Vector &u_fe = data.solid->u_fe;
 
       data.reset();
       scratch.reset();
       scratch.fe_values_ref.reinit(cell);
       cell->get_dof_indices(data.local_dof_indices);
+      const unsigned int n_independent_variables =
+          data.local_dof_indices.size();
+      const unsigned int n_dependent_variables = dofs_per_cell;
 
       const std::vector<std::shared_ptr<const PointHistory<dim,ADNumberType> > > lqph =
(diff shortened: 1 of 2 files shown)


jppelteret:
Currently WIP because I would like to generalise this to all `ad_type_code`s that we support (i.e. enable ADOL-C support and choose the AD library in the parameter file).

Quasi_static_Finite_strain_Compressible_Elasticity/cook_membrane.cc
```diff
@@ -1800,13 +1806,16 @@ namespace Cook_Membrane
...
-#ifdef ENABLE_SACADO_FORMULATION
+#ifdef ENABLE_AD_FORMULATION
 
+  namespace AD = dealii::Differentiation::AD;
 
   template <int dim>
   struct Assembler<dim,Sacado::Fad::DFad<double> > : Assembler_Base<dim,Sacado::Fad::DFad<double> >
   {
-    typedef Sacado::Fad::DFad<double> ADNumberType;
+    using ADHelper = AD::ADHelperResidualLinearization<AD::NumberTypes::sacado_dfad,double>;
```
davydden: I started applying this to the code derived from the code gallery and indeed, we certainly need a way to go from an arbitrary (AD) `NumberType` to the enumerator, and thus as you said, we need something like `none` for non-AD types.

jppelteret: I actually like the idea of an `AD::NumberTypes::None`. It would have simplified some of my own code, so I think I'd still like to implement it. It shouldn't be too much work.

davydden: sure, I agree, it will be useful regardless. What I meant is that even without `None` the link from AD numbers to enumerators can help.

jppelteret: For the record, `ADNumberTraits<ADNumberType>::type_code` defines this backwards link.

davydden:
> Currently WIP because I would like to generalise this to all ad_type_codes that we support (i.e. enable ADOL-C support and choose the AD library in the parameter file).

I played around and got the support for any AD numbers working (seems so at least) in my derived code using `AD::NumberTraits::n_supported_derivative_levels`. However this constraints the choice between residual and energy AD purely based upon `n_supported_derivative_levels` meaning one can't use `adolc_taped` for residual linearization. That's probably not how we should recommend using ADHelper, but that's how I got this setup for now. Just my 2c.

jppelteret:
> However this constraints the choice between residual and energy AD purely based upon n_supported_derivative_levels meaning one can't use adolc_taped for residual linearization.

You could get around this with a more specific check, for example:
```
enable_if<ADNumberTraits::is_taped || (ADNumberTraits::is_tapeless && ADNumberTraits::n_supported_derivative_levels >= 2)>
```
I think that this should work in your case.

PR #52, opened 2019-03-20
Add a code that solves a Bayesian inverse problem

bangerth:
Here is why I've been a bit absent over the past couple of weeks: I've actually gotten to program something :-)

This program solves a Bayesian inverse problem where we are sampling from a posterior probability in coefficient space in order to match as best as possible a previously determined "exact" set of measurements. The paper that describes the purpose of all of this is about to be written. It will details the results of solving more than 10^10 PDEs, using several CPU years.


tjhei:
Very cool! The setup looks somewhat familiar. :-)

I will look over the code in more detail soon.

bangerth:
Yes, that would be much appreciated. It's a first step towards working on a project you have something to do with as well :-)

My count for this morning: 8e9 samples already computed. I'll have a lot of data (terabytes) to evaluate by next week!

MCMC-Laplace/mcmc-laplace.cc
```diff
@@ -0,0 +1,723 @@
...
+    }
+
+    // And then define the tools to do point evaluation. We choose
+    // a set of 13x13 points evenly distributed across the domain:
+    {
+      const unsigned int n_points_per_direction = 13;
+      const double       dx = 1. / (n_points_per_direction + 1);
+
+      Vector<double>     weights(dof_handler.n_dofs());
+      FullMatrix<double> full_measurement_matrix(n_points_per_direction *
```
tjhei: move down to where you use it?

MCMC-Laplace/mcmc-laplace.cc
```diff
@@ -0,0 +1,723 @@
...
+
+    // Finally prebuild the building blocks of the linear system as
+    // discussed in the Readme file:
+    {
+      const unsigned int dofs_per_cell = fe.dofs_per_cell;
+
+      cell_matrix.reinit(dofs_per_cell, dofs_per_cell);
+      cell_rhs.reinit(dofs_per_cell);
+
+      const QGauss<dim>  quadrature_formula(2);
```
tjhei: compute degree based on fe degree?

MCMC-Laplace/mcmc-laplace.cc
```diff
@@ -0,0 +1,723 @@
...
+
+            cell_rhs(i) += (fe_values.shape_value(i, q_index) * // phi_i(x_q)
+                            10.0 *                              // f(x_q)
+                            fe_values.JxW(q_index));            // dx
+          }
+
+      VectorTools::interpolate_boundary_values(dof_handler,
+                                               0,
+                                               ZeroFunction<dim>(),
+                                               boundary_values);
```
tjhei: do you need to clear boundary_values?

MCMC-Laplace/mcmc-laplace.cc
```diff
@@ -0,0 +1,723 @@
...
+        this_cell_matrix = cell_matrix;
+        this_cell_matrix *= coefficients(cell->user_index());
+
+        cell->get_dof_indices(local_dof_indices);
+        for (unsigned int i = 0; i < dofs_per_cell; ++i)
+          {
+            for (unsigned int j = 0; j < dofs_per_cell; ++j)
+              system_matrix.add(local_dof_indices[i],
+                                local_dof_indices[j],
+                                this_cell_matrix(i, j));
```
tjhei: you could do the scaling in here and skip storing the scaled matrix.

MCMC-Laplace/mcmc-laplace.cc
```diff
@@ -0,0 +1,723 @@
...
+
+    for (const auto &cell : dof_handler.active_cell_iterators())
+      {
+        this_cell_matrix = cell_matrix;
+        this_cell_matrix *= coefficients(cell->user_index());
+
+        cell->get_dof_indices(local_dof_indices);
+        for (unsigned int i = 0; i < dofs_per_cell; ++i)
+          {
+            for (unsigned int j = 0; j < dofs_per_cell; ++j)
```
tjhei: you care about speed here, right? ``constraint_matrix.distribute_local_to_global()`` will be faster than a manual loop even if you have no constraints.

MCMC-Laplace/mcmc-laplace.cc
```diff
@@ -0,0 +1,723 @@
...
+              system_matrix.add(local_dof_indices[i],
+                                local_dof_indices[j],
+                                this_cell_matrix(i, j));
+
+            system_rhs(local_dof_indices[i]) += cell_rhs(i);
+          }
+      }
+
+    MatrixTools::apply_boundary_values(boundary_values,
```
tjhei: this is probably also slower than using a ConstraintMatrix.

MCMC-Laplace/mcmc-laplace.cc
```diff
@@ -0,0 +1,723 @@
...
+  //
+  // The function will also output the solution in a graphical format
+  // if you un-comment the corresponding statement in the third
+  // code block. However, you may end up with a very large amount
+  // of data: This code is producing, at the minimum, 10,000 samples
+  // and creating output for each one of them is surely more data
+  // than you ever want to see!
+  //
+  // At the end of the function, we output some timing information for
+  // the computations once for every 10,000 samples.
```
tjhei: remove "once"

MCMC-Laplace/mcmc-laplace.cc
```diff
@@ -0,0 +1,723 @@
...
+
+// The following namespaces define the statistical properties of the Bayesian
+// inverse problem. The first is about the definition of the measurement
+// statistics (the "likelihood"), which we here assume to be a normal
+// distribution $N(\mu,\sigma I)$ with mean value $\mu$ given by the
+// actual measurement vector (passed as an argument to the constructor
+// of the `Gaussian` class and standard deviation $\sigma$.
+//
+// For reasons of numerical accuracy, it is useful to not return the
+// actual likelihood, but it's logarithm. This is because these
```
tjhei: its

MCMC-Laplace/mcmc-laplace.cc
```diff
@@ -0,0 +1,723 @@
...
+
+  private:
+    ForwardSimulator::Interface &       simulator;
+    const LogLikelihood::Interface &    likelihood;
+    const LogPrior::Interface &         prior;
+    const ProposalGenerator::Interface &proposal_generator;
+
+    std::mt19937 random_number_generator;
+
+    unsigned int sample_number;
```
tjhei: maybe _count? or n_samples and n_accepted_samples?

MCMC-Laplace/mcmc-laplace.cc
```diff
@@ -0,0 +1,723 @@
...
+    Vector<double> current_sample = starting_guess;
+    double         current_log_posterior =
+      (likelihood.log_likelihood(simulator.evaluate(current_sample)) +
+       prior.log_prior(current_sample));
+
+    ++sample_number;
+    ++accepted_sample_number;
+    write_sample(current_sample, current_log_posterior);
+
+    for (unsigned int k = 1; k < n_samples; ++k, ++sample_number)
```
tjhei: why do you keep sample_number and k separate?

MCMC-Laplace/mcmc-laplace.cc
```diff
@@ -0,0 +1,723 @@
...
+            (std::exp(trial_log_posterior - current_log_posterior) >=
+             uniform_distribution(random_number_generator)))
+          {
+            current_sample        = trial_sample;
+            current_log_posterior = trial_log_posterior;
+
+            ++accepted_sample_number;
+          }
+
+        write_sample(current_sample, current_log_posterior);
```
tjhei: does it make sense to write out samples you won't accept?

MCMC-Laplace/mcmc-laplace.cc
```diff
@@ -0,0 +1,723 @@
...
+
+
+
+  template <int dim>
+  PoissonSolver<dim>::PoissonSolver(const unsigned int global_refinements,
+                                    const unsigned int fe_degree,
+                                    const std::string &dataset_name)
+    : fe(fe_degree)
+    , dof_handler(triangulation)
+    , timer(std::cout, TimerOutput::summary, TimerOutput::cpu_times)
```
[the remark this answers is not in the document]
bangerth: Mostly CPU times. I'm running this thing on machines in parallel, so I was mostly interested in how many milliseconds per PDE solve. (The answer is below 4ms.)

MCMC-Laplace/mcmc-laplace.cc
bangerth: Not really -- it's the same value we write into the same location of the `std::map` every time. So it may actually be cheaper to not clear the map, but it's probably good practice. So I'll do that.

bangerth: Actually, this is in the `setup_system()` function that is called only once, from the constructor. So there is no prior content in the variable.

MCMC-Laplace/mcmc-laplace.cc
bangerth: Tricky. When we have functions that are called `n_...()`, I think of this as a container and I'm asking for the number of elements it stores. So that would mean that I can access these samples, but that's not the case here: No way to store all samples, they only exist on disk. I'm really just thinking of these variables as the "how manyth sample are we currently processing", not as "how many are there that one can access.

So I'd rather leave the name as is.

MCMC-Laplace/mcmc-laplace.cc
bangerth: The function can be called multiple times. Each time, it computes `n_samples`, and for each sample produced, it has to have a local counter (`k`) and a variable that keeps track how many samples this has added up over the course of all invocations of the function.

MCMC-Laplace/mcmc-laplace.cc
bangerth: Oh, most definitely! MH only works if you keep (repeat) the previous sample every time you discard the current one. That's how the algorithm works :-)

tjhei: oh, this is ``current_sample`` and not the rejected one. Oops.

MCMC-Laplace/mcmc-laplace.cc
tjhei: You are not doing that here but I guess that makes sense to allow.

bangerth:
I'm making the changes you suggested, except the switch to the `AffineConstraints` for both `distribute_local_to_global` and to avoid the call to `MatrixTools::apply_b_v()`. I tried this out and it makes the assembly take twice as much time, so I'm sticking with my existing code :-)

bangerth:
By way of reference, assembly only takes 10% of the total time after all of the optimizations I'm already doing, so I'm not terribly concerned about that part of the program. It may be worthwhile reconsidering after you add a variation of this program to the code gallery that uses matrix-free solvers :-)

tjhei:
> I tried this out and it makes the assembly take twice as much time

That is quite interesting (and unexpected)!

> assembly only takes 10% of the total time

I am surprised assembly is so cheap.

> It may be worthwhile reconsidering after you add a variation of this program to the code gallery that uses matrix-free solvers :-)

Okay. :-)

bangerth:
On 3/27/19 8:05 PM, Timo Heister wrote:
>     I tried this out and it makes the assembly take twice as much time
> 
> That is quite interesting (and unexpected)!

It's because I don't have any hanging node constraints. AffineConstraints is 
written for the complicated case, but I really only have equality constraints.


>     assembly only takes 10% of the total time
> 
> I am surprised assembly is so cheap.

And that's because I really don't do anything. What's expensive is the 
reinit() of FEValues, and the triple loop over i, j, q. But I have neither -- 
it's all pre-computed.

tjhei:
Looks good. Is this ready to go @bangerth ?

bangerth:
No, I still need to deal with the issue of the wrong quadrature formula. That's a bummer because I had computed the "exact" data using the wrong quadrature. I checked this this morning and the numbers I used are *almost* as one gets with the right quadrature formula, but not quite. I could ignore this, or I could just run all 16B samples again -- I think I'll do the latter. 3 CPU years flushed down the toilet...

bangerth:
OK, ready now. This relies on https://github.com/dealii/dealii/pull/7871 in the last commit, but I think it is ok to let code-gallery programs rely on the deal.II dev version, right?

MCMC-Laplace/mcmc-laplace.cc
```diff
@@ -661,7 +661,32 @@ namespace Sampler
 
 
 // The final function is `main()`, which simply puts all of these pieces
-// together into one:
+// together into one. The "exaxt solution", i.e., the "measurement values"
```
tjhei: typo "exact"

MCMC-Laplace/mcmc-laplace.cc
```diff
@@ -675,23 +700,92 @@ int main()
   const unsigned int random_seed  = (testing ? 1U : std::random_device()());
   const std::string  dataset_name = std::to_string(random_seed);
 
-  // Set the exact coefficient:
```
tjhei: why not keep this code in an ``if (false)`` block or something?

tjhei:
> OK, ready now. This relies on [dealii/dealii#7871](https://github.com/dealii/dealii/pull/7871) in the last commit, but I think it is ok to let code-gallery programs rely on the deal.II dev version, right?

I guess so? I think you should change your CMakeList to request ``>=9.1.0.pre`` though.

MCMC-Laplace/mcmc-laplace.cc
bangerth: I'm just not a fan of leaving dead code around. I like it better in the comment.

bangerth:
OK, all updated.

tjhei:
> 3 CPU years flushed down the toilet...

That's annoying but not as much as I have wasted in the last few weeks. ;-)
I guess we are all contributing to global warming in some way. :-)

MCMC-Laplace/mcmc-laplace.cc
tjhei: I prefer dead code that at least continues to compile over dead code in comments.

I guess we disagree on that front, but fine with me. :-)

bangerth:
Now I'm curious: What did you do?

tjhei:
> Now I'm curious: What did you do?

Doing the large runs on stampede2 with 20k+ cores with @tcclevenger and discovering all sorts of problems.

bangerth:
Alright -- first new code gallery code in a while!

20k+ cores -- yes, CPU time goes fast at that scale :-) Could you have foreseen the problems?

PR #55, opened 2019-05-21
cdr: Update to 9.1

tamiko:


tamiko:
In reference to https://github.com/dealii/dealii/issues/8080

cdr/solver/cdr.cc
```diff
@@ -232,9 +232,12 @@ void CDRProblem<dim>::time_iterate()
 template<int dim>
 void CDRProblem<dim>::refine_mesh()
 {
+  using FunctionMap =
+    std::map<types::boundary_id, const Function<dim> *>;
```
tamiko: @bangerth *jup*

bangerth: I see. Thanks!

PR #56, opened 2019-05-21
CeresFE: Update to 9.1

tamiko:


tamiko:

CeresFE/src/ceres.cc
```diff
@@ -1996,12 +1995,13 @@ namespace Step22
   template<int dim>
   void StokesProblem<dim>::refine_mesh()
   {
+    using FunctionMap = std::map<types::boundary_id, const Function<dim> *>;
```
bangerth: same question here

PR #58, opened 2019-05-21
ElastoplasticTorsion: Update to 9.1

tamiko:


tamiko:

ElastoplasticTorsion/ElastoplasticTorsion.cc
```diff
@@ -763,10 +763,12 @@ namespace nsp
   template <int dim>
   void ElastoplasticTorsion<dim>::refine_mesh ()
   {
+    using FunctionMap = std::map<types::boundary_id, const Function<dim> *>;
+
```
bangerth: Why do you need this? Don't we define this `typedef` in deal.II already?

tamiko: I don't think so. The `FunctionMap` class template that had a `FunctionMap<dim>::type` declaration is deprecated. The corresponding Kelly error estimator functions use above `std::map` directly. Personally, I think that the `using FunctionMap = ...` is slightly more readible.

bangerth: Did I see this wrong in the other PRs I just merged that you just used a bar `FunctionMap()` argument? Or was there a `using` declaration as well that I missed?

tamiko: There was a `using` declaration as well :-)

PR #60, opened 2019-06-12
Add placeholder for Magneto-active polymers book

jppelteret:


jppelteret:
> Are you afraid someone steals the directory name?

Err, no :-D I'm providing this link in the book, so it'd be a crying shame if it didn't work. I'd prefer to have the link in place and test it before submission. Is that ok?

bangerth:
Good argument!

PR #61, opened 2019-06-21
Avoid deprecated stuff.

bangerth:


bangerth:
While there also indent member variables.

goal_oriented_elastoplasticity/elastoplastic.cc
```diff
@@ -2395,37 +2394,37 @@ namespace ElastoPlastic
     void output_results ();
 
     const FESystem<dim>       &fe;
-    DoFHandler<dim>           dof_handler;
-    const Vector<double>      solution;
+    DoFHandler<dim>            dof_handler;
```
tjhei: so this means we are not running clang-format on the code gallery? Maybe that would be worth adding...

bangerth: Correct. We have always left formatting up to the authors of these programs, and I think that's the right approach (given that we only *distribute* these programs, not claim ownership or provide any kind of guarantee for their correctness).

PR #63, opened 2019-06-30
Thermo-electro-elastic pump

jppelteret:
Implementation of a problem with thermo-electro-mechanical coupling

The electro-mechanical and thermal problems are solved in a staggered manner.
Two implementations of the electro-mechanical problem are implemented:
- the standard two-field formulation using the electric scalar potential and a
  displacement field (valid for compressible materials), and
- a four-field formulation, extending the above with an additional dilatation
  and pressure field (valid for both compressible and near-incompressible 
  materials).


jppelteret:
This is still a WIP, so please do not review or merge this yet.

PR #64, opened 2020-01-02
Fix a bug in the Metropolis-Hastings solver.

bangerth:
This is rather a big nuisance: I created 10^11 samples with the existing program, burning ~15 CPU years. But it turns out that these samples can all go into the trash: The sampling algorithm had a subtle bug :-(

### DIFF
--- a/MCMC-Laplace/mcmc-laplace.cc
+++ b/MCMC-Laplace/mcmc-laplace.cc
@@ -842,7 +842,7 @@ int main()
   LogLikelihood::Gaussian        log_likelihood(exact_solution, 0.05);
   LogPrior::LogGaussian          log_prior(0, 2);
   ProposalGenerator::LogGaussian proposal_generator(
-    random_seed, 0.0725); /* so that the acceptance ratio is ~0.3 */
+    random_seed, 0.09); /* so that the acceptance ratio is ~0.24 */
   Sampler::MetropolisHastings sampler(laplace_problem,
                                       log_likelihood,
                                       log_prior,
--- a/MCMC-Laplace/mcmc-laplace.cc
+++ b/MCMC-Laplace/mcmc-laplace.cc
@@ -74,6 +74,8 @@ namespace ForwardSimulator
     virtual Vector<double> evaluate(const Vector<double> &coefficients) = 0;
   };
 
+
+
   template <int dim>
   class PoissonSolver : public Interface
   {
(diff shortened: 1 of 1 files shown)


MCMC-Laplace/mcmc-laplace.cc
```diff
@@ -499,12 +501,26 @@ namespace LogPrior
...
+// component the ratio of these probabilities is $y_i/x_i$, and
+// consequently for all components of the vector together, the
+// probability is the product of these ratios.
 namespace ProposalGenerator
 {
   class Interface
   {
   public:
-    virtual Vector<double>
+    virtual
```
tjhei:
```suggestion
    /**
     * Draw a new sample based on the current sample @p current_sample.
     *
     * This functions returns a pair consisting of the new sample and the ratio of the probability
     * for reaching the new from the old divided by the probability of reaching the old from the new.
     */
    virtual
```

MCMC-Laplace/mcmc-laplace.cc
```diff
@@ -624,14 +649,22 @@ namespace Sampler
...
         const double trial_log_posterior =
           (likelihood.log_likelihood(simulator.evaluate(trial_sample)) +
            prior.log_prior(trial_sample));
 
-        if ((trial_log_posterior > current_log_posterior) ||
-            (std::exp(trial_log_posterior - current_log_posterior) >=
+        if ((trial_log_posterior + std::log(perturbation_probability_ratio)
+             >=
+             current_log_posterior)
+            ||
```
tjhei: While this looks correct (well, compared to wikipedia at least), I am not sure why you do only one of the comparisons in log space. Does it matter for accuracy which one to pick?

tjhei:
> The sampling algorithm had a subtle bug

Oh, no. That is unfortunate. :-(

MCMC-Laplace/mcmc-laplace.cc
bangerth: I wanted to avoid this kind of documentation style. The code gallery programs are processed like the tutorials, not like the library. That is, we don't document individual functions, and furthermore, `/* ... */` are just copied verbatim into the output. So I would like to stick with the style I chose to document.

MCMC-Laplace/mcmc-laplace.cc
bangerth: The intent of the comparison in the original code was that if the trial probability is larger than the previous probability, then this can be tested quickly by just comparing the previously computed logs of the two. Only if that test fails do we get into the other branch of the `||` statement that then requires the computation of the (expensive) exponential. This continues to be true: With the new code, the first branch has only one `log`. We could remove the first branch altogether, but then have to do two `exp` operations.

tjhei: Does the performance of the exp or log in this check really matter? I assume not because we make thousands of operations to pick a new sample, so the clearest layout makes the most sense. From your answer it sounds like accuracy does not matter here.

bangerth: Probably not.

bangerth: But it is also the random number generation that is bypassed this way: If we accept a sample early, we don't call the rng.

That said, that's likely also a minor cost, at least for the current program.

bangerth:
@tjhei : I've updated it to your wishes (with the exception of the function `/** ... */` comment, which we generally don't do in code gallery programs. I've also pushed a couple of other minor adjustments.

MCMC-Laplace/mcmc-laplace.cc
```diff
@@ -499,12 +501,26 @@ namespace LogPrior
 // to multiply the existing sample entries by are close to one. And
 // because the exponential of a number is always positive, we never
 // get negative samples this way.)
+//
+// But the Metropolis-Hastings sampler doesn't just need a perturbed
+// sample $y$ location given the current sample location $x$. It also
+// needs to know the ratio of the probability of reaching $y$ from
+// $x$, divided by the probability of reaching $x$ from $y$. If we
```
tjhei: Just to be sure: Are you taking the correct ratio here? The acceptance ratio A uses g(x|y)/g(y|x) to decide if we should accept y (https://en.wikipedia.org/wiki/Metropolis%E2%80%93Hastings_algorithm#Formal_derivation).

bangerth: Yes, I had double checked it by running it this way and with the reverse ratio, and this was correct.

PR #82, opened 2020-06-24
Removes a redundant edge from Nonlinear_PoroViscoelasticity tutorial dependencies

krishnakumarg1984:


krishnakumarg1984:
Thanks for merging

jppelteret:
Thanks for the fix. The original list better reflected how we went about developing this code, but the amendment makes sense in terms of what we're trying to show in the tutorial graph.

krishnakumarg1984:
I see. I didn't know about the background. I just thought that the tutorial flow-diagram looked a bit messy :)

jppelteret:
And I haven't seen what the graph looked like before the patch was merged :-D As I said, it felt right to add all of those entries in when we documented the code, but I think that your changes are ultimately correct and in the spirit of what information we're trying to transmit to everyone.

PR #84, opened 2020-10-26
Add a deal.II-preCICE example

DavidSCN:
...following the discussion in https://github.com/dealii/dealii/issues/10934.

The tutorial consists of a time-dependent Laplace problem and an externally coupled time-dependent C++ boundary condition. I think it is best to not add further information here, everything should be explained in the source code/ README file. The only thing to discuss is how to treat the dependency preCICE in the `dependencies` file, since it is usually refers to dea.II compile dependencies. I currently added just 'preCICE' as user information. Any feedback is of course very welcome!

Resolves https://github.com/dealii/dealii/issues/10934.


coupled_laplace_problem/coupled_laplace_problem.cc
```diff
@@ -0,0 +1,638 @@
...
+  SparseMatrix<double> system_matrix;
+
+  Vector<double> solution;
+  Vector<double> old_solution;
+  Vector<double> system_rhs;
+
+  // We allocate all structures required for the preCICE coupling. The map
+  // is used to apply Dirichlet boundary conditions and filled in the Adapter
+  // class with data from the other participant. The CouplingParameters hold the
+  // preCICE configuration as described above. The interdace boundary ID is the
```
peterrum, line 351: interdace -> interface

coupled_laplace_problem/coupled_laplace_problem.cc
```diff
@@ -0,0 +1,638 @@
...
+  for (const auto &cell : triangulation.active_cell_iterators())
+    for (auto f : GeometryInfo<dim>::face_indices())
+      {
+        const auto face = cell->face(f);
+
+        // We choose the boundary in positive x direction for the
+        // interface coupling.
+        if (face->at_boundary() && (face->center()[0] == 1))
+          face->set_boundary_id(interface_boundary_id);
+      }
```
peterrum, line 433: Why don't you pass to the `hyper_cube` function a bool to color the faces?

coupled_laplace_problem/coupled_laplace_problem.cc
```diff
@@ -0,0 +1,638 @@
...
+      cell->get_dof_indices(local_dof_indices);
+      for (const unsigned int i : fe_values.dof_indices())
+        {
+          for (const unsigned int j : fe_values.dof_indices())
+            system_matrix.add(local_dof_indices[i],
+                              local_dof_indices[j],
+                              cell_matrix(i, j));
+
+          system_rhs(local_dof_indices[i]) += cell_rhs(i);
+        }
```
peterrum, line 534: Why don't you use `AffineConstraints::distribute_local_to_global`?

coupled_laplace_problem/coupled_laplace_problem.cc
```diff
@@ -0,0 +1,638 @@
...
+template <int dim>
+void
+CoupledLaplaceProblem<dim>::output_results() const
+{
+  DataOut<dim> data_out;
+
+  data_out.attach_dof_handler(dof_handler);
+  data_out.add_data_vector(solution, "solution");
+
+  data_out.build_patches();
```
peterrum, line 584: Could you use a global mapping in the example (and not rely on the default values).

coupled_laplace_problem/coupled_laplace_problem.cc
```diff
@@ -0,0 +1,638 @@
...
+      // In the time loop, we assemble the coupled system and solve it as
+      // usual.
+      assemble_system();
+      solve();
+
+      // After we solved the system, we advance the coupling to the next time
+      // level. In a bi-directional coupled simulation, we would pass our
+      // calculated data to and obtain new data from preCICE. Here, we simply
+      // obtain new data from preCICE, so from the other participant.
+      delta_t = adapter.advance(boundary_data, delta_t);
```
peterrum, line 621: `delta_t` as input and output is a bit confusing.

coupled_laplace_problem/coupled_laplace_problem.cc
```diff
@@ -0,0 +1,638 @@
...
+
+      // Write output file if time step is complete
+      if (adapter.precice.isTimeWindowComplete())
+        output_results();
```
peterrum, line 625: Does that mean that there might be sub-timesteps?

peterrum (commented):
Some comments and questions.

I would suggest to add a GitHub action to test the new program. In the action, one could use a specified deal.II docker image from https://hub.docker.com/r/dealii/dealii/ (e.g. 9.2). Inside, one could install all the needed dependencies (I will post a draft later today). The same steps performed there could be also mentioned in the readme, which should be extended with how to run the program.

peterrum:
Maybe you could add a GitHub-action like this (modified so that it also installs all the dependencies):
```
name: PreCICE CI

on:
  push:
    branches: [ master ]
  pull_request:
    branches: [ master ]

jobs:
  build:

    runs-on: ubuntu-latest

    steps:
    - uses: actions/checkout@v2
    - name: setup
      run: |
         command="git clone https://github.com/${{ github.repository }} && \
                  cd code-gallery && \
                  git fetch origin ${{ github.ref }} && \
                  git checkout FETCH_HEAD && \
                  cd coupled_laplace_problem && \
                  cmake . && \
                  make && \
                  ctest";
         
         echo $command
         
         docker pull dealii/dealii:master-focal
         docker run -t dealii/dealii:master-focal /bin/sh -c "$command";
```
Instead of `dealii/dealii:master-focal` you should select the image of 9.1 or 9.2.

The posted implementation is probably not the nicest implementation. Maybe you have a better solution. Maybe https://www.petefreitag.com/item/903.cfm helps.

coupled_laplace_problem/coupled_laplace_problem.cc
```diff
@@ -0,0 +1,638 @@
...
+  DoFTools::map_dofs_to_support_points(MappingQ1<dim>(),
+                                       dof_handler,
+                                       support_points);
+
+  // `support_points` contains now the coordinates of all DoFs. In the next
+  // step, the relevant coordinates are extracted using the IndexSet with the
+  // extracted coupling_dofs.
+
+  // preCICE expects all data in the format [x0, y0, z0, x1, y1 ...]. In 2D, the
+  // z coordinate needs to be omitted.
```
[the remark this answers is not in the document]
DavidSCN: I agree, but a selective computation is not provided in deal.II (to my best knowledge). Since the utilization in parallel (not relevant for this tutorial, but for the other codes) is also not possible with this function I might find a solution for both problems looking at [the discussed code snippet](https://github.com/dealii/dealii/blob/ec916371d189c057adeb0e15d0f2d1a1982a4d87/include/deal.II/numerics/vector_tools_interpolate.templates.h#L299-L320).

coupled_laplace_problem/coupled_laplace_problem.cc
```diff
@@ -0,0 +1,638 @@
...
+  int node_position_iterator = 0;
+  for (auto element : coupling_dofs)
+    {
+      for (int i = 0; i < dim; ++i)
+        interface_nodes_positions[node_position_iterator * dim + i] =
+          support_points[element][i];
```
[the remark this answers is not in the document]
DavidSCN: The vector is directly allocated in the appropriate size. So, I can even skip the `reserve`.

coupled_laplace_problem/coupled_laplace_problem.cc
DavidSCN: I need to summarize all non-coupled boundaries in a single ID in order to apply the 'usual' DBC. Hence, there is not really a difference in reassigning the colored boundaries or assigning the coupling boundary.

coupled_laplace_problem/coupled_laplace_problem.cc
DavidSCN: Mainly due to the [tutorial implementation](https://github.com/dealii/dealii/blob/612d755edcfe45b588fabc9b2f298ce57f15eacc/examples/step-4/step-4.cc#L411-L421). I haven't changed these lines. Should I do it anyway?

coupled_laplace_problem/coupled_laplace_problem.cc
DavidSCN: It depends on the configuration: the function prevents writing results in case of an implicit coupling, where a time step is computed several times until convergence. Sub-timesteps of individual participants are also possible (this is called 'subcycling'). We don't have any of these methods in the default tutorial configuration (and the solver code is not really prepared for them). I added a comment.

coupled_laplace_problem/coupled_laplace_problem.cc
```diff
@@ -0,0 +1,638 @@
...
+// computation in each time step. Here,
+// coupling data is passed to and obtained from preCICE.
+template <int dim, typename ParameterClass>
+double
+Adapter<dim, ParameterClass>::advance(
+  std::map<types::global_dof_index, double> &boundary_data,
+  const double                               computed_timestep_length)
+{
+  // Here, we specify the computed time step length and pass it to preCICE
+  const double max_delta_t = precice.advance(computed_timestep_length);
```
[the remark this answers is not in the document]
DavidSCN: I added a comment. Let me know, if is is not sufficient.

coupled_laplace_problem/coupled_laplace_problem.cc
```diff
@@ -0,0 +1,638 @@
...
+// In addition to the deal.II header files, we include the preCICE API in order
+// to obtain access to preCICE specific functionality
+#include <precice/SolverInterface.hpp>
+
+#include <fstream>
+#include <iostream>
+
+using namespace dealii;
+
+// @sect3{Configuration parameters}
```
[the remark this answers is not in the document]
DavidSCN: I'm not even sure, if the @sect tags have any influence. I realized that most of the code-gallery examples don't have any of them. I guess I can remove them completely?

coupled_laplace_problem/coupled_laplace_problem.cc
DavidSCN: I will try to find every place.

coupled_laplace_problem/coupled_laplace_problem.cc
```diff
@@ -0,0 +1,638 @@
...
+  // preCICE related initializations
+  // These variables are specified in and read from a parameter file, which is
+  // in this simple tutorial program the CouplingParameter struct already
+  // introduced in the beginning.
+  const std::string mesh_name;
+  const std::string read_data_name;
+
+  // These IDs are filled by preCICE during the initialization. We set a default
+  // value of -1 in order to detect potential errors more easily.
+  int mesh_id           = -1;
```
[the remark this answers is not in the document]
DavidSCN: precice uses `int` instead of `unsigned int`. I'm not completely sure about the reason, probably due compatibility to other languages. It is probably not possible to obtain negative IDs, but I would like to keep the variable type consistently (no unsigned). I think the compiler might throw warnings otherwise. I removed the default arguments completely.

coupled_laplace_problem/fancy_boundary_condition.cc
```diff
@@ -0,0 +1,107 @@
...
+int
+main()
+{
+  std::cout << "Boundary participant: starting... \n";
+
+  // Adjust to MPI rank and size for parallel computation
+  const int commRank = 0;
+  const int commSize = 1;
+
+  using namespace precice;
```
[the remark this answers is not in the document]
DavidSCN: What do you mean? Adding a comment why I use it here or prefixing every use of a precice related function?

coupled_laplace_problem/coupled_laplace_problem.cc
DavidSCN: I made the whole procedure now much more verbose. The input output confusion is in any case resolved.

coupled_laplace_problem/fancy_boundary_condition.cc
```diff
@@ -0,0 +1,107 @@
...
+
+
+int
+main()
+{
+  std::cout << "Boundary participant: starting... \n";
+
+  // Adjust to MPI rank and size for parallel computation
+  const int commRank = 0;
+  const int commSize = 1;
```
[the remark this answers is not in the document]
DavidSCN: It's not planned and we have no default arguments. Still, it should make things more clear if we write out the variable names instead of using hard-coded numbers. Or do you have any particular reason to remove it?

coupled_laplace_problem/fancy_boundary_condition.cc
```diff
@@ -0,0 +1,107 @@
...
+  const double relative_time = time / end_time;
+  // Define the amplitude. Values run from -0.5 to 0.5
+  const double amplitude = (relative_time - 0.5);
+  // Specify the actual data we want to pass to the other participant. Here, we
+  // choose a parabola with boundary values 2 in order to enforce continuity
+  // to adjacent boundaries.
+  const unsigned int n_elements = boundary_data.size();
+  for (uint i = 0; i < n_elements; ++i)
+    boundary_data[i] = amplitude * (i * ((n_elements - 1) - i)) + 2;
+}
```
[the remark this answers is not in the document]
DavidSCN: I can not do it in the deal.II manner, because the code doesn't use deal.II. I tried to make it more clear.

DavidSCN:
Thanks for the review @peterrum. I still need to adjust some things and will re-request afterwards..
Testing is a good idea. I'm not a specialist in GitHub actions, but precice can (hopefully) directly be installed using the debian binaries. I added the tests to my TODO list and will address the remaining tasks throughout this week:
- [ ] extend and adjust README (see review comments)
- [ ] add CI using GitHub actions
- [ ] find a solution for boundary node mapping (`map_dofs_to_support_points()`)

DavidSCN:
So, the Docker command should be complete. But the executables print `vtk` files. Should I manually add some reference and add a comparison with `diff` ?

peterrum:
> Should I manually add some reference and add a comparison with diff ?

Sounds good!

coupled_laplace_problem/README.md
```diff
@@ -0,0 +1,67 @@
...
+@f{align*}
+  \\frac{u^{n+1}-u^n}{\Delta t}-\Delta u^{n+1} &= f \qquad\qquad & \text{in}\ \Omega,
+@f}
+at the next time level n+1.
+
+## Requirements
+
+* `deal.II`, version `9.2` or greater. Older versions might work as well, but have not been tested.
+
+* [preCICE](https://github.com/precice/precice/wiki#1-get-precice), version `2.0` or greater.
```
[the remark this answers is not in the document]
uekerman: There are different options depending on the system. Giving any option here explicitly could lead to the misunderstanding that it is the only way to go. Giving more options would in the end lead to a duplication of the content of the linked page. Just having that link could be the best compromise. What do you think?

coupled_laplace_problem/fancy_boundary_condition.cc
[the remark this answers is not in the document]
uekerman: The latter. I see the point. In the end you only need to add the namespace to `SolverInterface`.

coupled_laplace_problem/README.md
```diff
@@ -0,0 +1,67 @@
+Laplace equation coupled to an external simulation program
+------------------------------------------
+## Overview
+
+preCICE allows to couple deal.II to external simulation software, such as OpenFOAM, SU2, or CalculiX. To keep dependencies of this example minimal, we couple deal.II to an external C++ program, which provides a time varying boundary condition. The deal.II code consists mainly of the [`step-4` tutorial program](https://www.dealii.org/developer/doxygen/deal.II/step_4.html), where a simple Laplace problem is solved. 
+
+Coupling with preCICE is usually carried out along surfaces in order to apply a Dirichlet-Neumann coupling between two domains (volume coupling is also possible). For the sake of simplicity, we couple the external C++ program in a unidirectional fashion to one side of our quadrilateral domain. The external C++ program generates a parabolic boundary profile with time varying amplitude. The boundary values are then used in the Laplace solver as a Dirichlet boundary condition.
+
+## Time discretization
+Coupled simulations deal mostly with time-dependent problems. Hence, we modify the stationary Laplace problem from step-4 by a time-dependent problem
```
uekerman, line 10:
```suggestion
Coupled simulations deal mostly with time-dependent problems. Hence, we make the stationary Laplace problem from step-4 time dependent.
```

coupled_laplace_problem/fancy_boundary_condition.cc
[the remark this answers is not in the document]
DavidSCN: Done

DavidSCN:
Concerning your comment @peterrum 
> I guess this is not an optimal approach. You compute the coordinate of the support points in the whole domain but one only needs them on some faces.

I just adapted (draft) the `map_dofs_to_support_points`  function, but it becomes quite lengthy and I'm not sure if the efficiency gain at this point is too complicated. Maybe we can add this functionality to our existing tutorials. I still have  the issue to extract the `local_face_indices` instead of the whole cell indices. Opinions?
```cpp
template <int dim, int spacedim>
void
map_boundary_dofs_to_support_points(
  const Mapping<dim, spacedim> &                      mapping,
  const DoFHandler<dim, spacedim> &                   dof_handler,
  std::map<types::global_dof_index, Point<spacedim>> &support_points,
  const ComponentMask &                               in_mask,
  const types::boundary_id                            boundary_id)
{
  const FiniteElement<dim, spacedim> &fe = dof_handler.get_fe();
  // check whether every fe in the collection has support points
  Assert(fe.has_support_points(),
         typename FiniteElement<dim>::ExcFEHasNoSupportPoints());

  Quadrature<dim - 1> quad(fe.get_unit_face_support_points());

  // Take care of components
  const ComponentMask mask =
    (in_mask.size() == 0 ? ComponentMask(fe.n_components(), true) : in_mask);

  // Now loop over all cells and enquire the support points on each
  // of these. we use dummy quadrature formulas where the quadrature
  // points are located at the unit support points to enquire the
  // location of the support points in real space.
  //
  // The weights of the quadrature rule have been set to invalid
  // values by the used constructor.
  FEFaceValues<dim, spacedim>          fe_values(mapping,
                                        fe,
                                        quad,
                                        update_quadrature_points);

  std::vector<types::global_dof_index> local_dof_indices;
  for (const auto &cell : dof_handler.active_cell_iterators())
    for (const auto &face : cell->face_iterators())
      if (face->at_boundary() == true && face->boundary_id() == boundary_id &&
          cell->is_artificial() == false)
        // only work on locally relevant cells
        {
          fe_values.reinit(cell, face);

          local_dof_indices.resize(fe.dofs_per_cell);
          cell->get_dof_indices(local_dof_indices);

          const std::vector<Point<spacedim>> &points =
            fe_values.get_quadrature_points();
          // TODO: Add if statement for face DoFs
          for (unsigned int i = 0; i < fe.n_dofs_per_cell(); ++i)
            {
              const unsigned int dof_comp =
                fe.system_to_component_index(i).first;

              // insert the values into the map if it is a valid component
              if (mask[dof_comp])
                support_points[local_dof_indices[i]] = points[i];
            }
        }
}
```

DavidSCN:
@uekerman Running the case in docker fails due to an [RBF failure](https://github.com/DavidSCN/code-gallery/runs/1367174292?check_suite_focus=true#step:3:1260). I cannot reproduce it locally. Apart from that, the CI is ready.  Quick workaround would be to use another mapping config.

coupled_laplace_problem/Allclean
```diff
@@ -0,0 +1,25 @@
+#!/bin/sh
```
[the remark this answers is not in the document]
DavidSCN: Done

peterrum:
@DavidSCN Requesting a review on Saturday - just like back in the days :P

I'll have another round through the code in the next day.

coupled_laplace_problem/precice-config.xml
```diff
@@ -0,0 +1,43 @@
...
+
+    <mesh name="boundary-mesh">
+        <use-data name="boundary-data"/>
+    </mesh>
+
+    <participant name="laplace-solver">
+        <use-mesh name="dealii-mesh" provide="yes"/>
+        <use-mesh name="boundary-mesh" from="boundary-participant"/>
+        <read-data name="boundary-data" mesh="dealii-mesh"/>
+        <mapping:rbf-thin-plate-splines direction="read" from="boundary-mesh" to="dealii-mesh" constraint="consistent" x-dead="true"/>
```
uekerman, line 25:
```suggestion
        <mapping:rbf-thin-plate-splines direction="read" from="boundary-mesh" to="dealii-mesh" constraint="consistent" x-dead="true" use-lu-decomposition="true"/>
```

uekerman: I opened an issue https://github.com/precice/precice/issues/904
But let's not make this a show stopper here.

DavidSCN:
Test is now completed and runs successfully.

coupled_laplace_problem/coupled_laplace_problem.cc
```diff
@@ -0,0 +1,652 @@
...
+  // is read/written per interface node.
+  read_data.resize(n_interface_nodes);
+  // The IDs are again filled by preCICE during the initializations.
+  interface_nodes_ids.resize(n_interface_nodes);
+
+  // The node location is obtained using `map_dofs_to_support_points()`.
+  std::map<types::global_dof_index, Point<dim>> support_points;
+
+  // We use here a simple Q1 mapping. In case one has more complex
+  // geomtries, you might want to change this to a higher order mapping.
```
peterrum, line 214: This comment seems to be outdated? I suggest that you remove it.

coupled_laplace_problem/coupled_laplace_problem.cc
```diff
@@ -0,0 +1,652 @@
...
+
+  // `support_points` contains now the coordinates of all DoFs. In the next
+  // step, the relevant coordinates are extracted using the IndexSet with the
+  // extracted coupling_dofs.
+  for (auto element : coupling_dofs)
+    for (int i = 0; i < dim; ++i)
+      interface_nodes_positions.push_back(support_points[element][i]);
+
+  for (uint i = 0; i < interface_nodes_positions.size(); ++i)
+    std::cout << i << "    " << interface_nodes_positions[i] << std::endl;
```
peterrum, line 225: Do you really want to output the points to terminal? Or was this only for debugging?

coupled_laplace_problem/coupled_laplace_problem.cc
```diff
@@ -0,0 +1,652 @@
...
+  // the second participant in a serial coupling scheme)
+  if (precice.isReadDataAvailable())
+    {
+      precice.readBlockScalarData(read_data_id,
+                                  n_interface_nodes,
+                                  interface_nodes_ids.data(),
+                                  read_data.data());
+
+      // This is the opposite direction as above. See comment there.
+      format_precice_to_dealii(boundary_data);
```
peterrum, line 249: Which comment are you referring to?

coupled_laplace_problem/coupled_laplace_problem.cc
```diff
@@ -0,0 +1,652 @@
...
+  MappingQ1<dim>     mapping;
+
+  SparsityPattern      sparsity_pattern;
+  SparseMatrix<double> system_matrix;
+
+  Vector<double> solution;
+  Vector<double> old_solution;
+  Vector<double> system_rhs;
+
+  // We allocate all structures required for the preCICE coupling. The map
```
peterrum, line 346: // All structures required for the preCICE coupling: The map

coupled_laplace_problem/coupled_laplace_problem.cc
```diff
@@ -0,0 +1,652 @@
...
+
+template <int dim>
+void
+CoupledLaplaceProblem<dim>::make_grid()
+{
+  GridGenerator::hyper_cube(triangulation, -1, 1);
+  triangulation.refine_global(4);
+
+  for (const auto &cell : triangulation.active_cell_iterators())
+    for (auto f : GeometryInfo<dim>::face_indices())
```
peterrum, line 430:
```suggestion
    for (const auto face : cell->face_iterators())
```
Consider `GeometryInfo` as deprecated ;)

coupled_laplace_problem/coupled_laplace_problem.cc
```diff
@@ -0,0 +1,652 @@
...
+CoupledLaplaceProblem<dim>::run()
+{
+  std::cout << "Solving problem in " << dim << " space dimensions."
+            << std::endl;
+
+  make_grid();
+  setup_system();
+
+  // After we set up the system, we initialize preCICE using the functionalities
+  // of the Adapter. precice returns the maximum admissible time-step size,
```
peterrum, line 608:
```suggestion
  // of the Adapter. preCICE returns the maximum admissible time-step size,
```

peterrum (approved):
Looks good to me. Thanks! I have left some minor comments. Once these are addressed, it is ready from my side. 

Maybe someone else has some comments.  I let someone else merge!

coupled_laplace_problem/coupled_laplace_problem.cc
DavidSCN: Debug... I removed it.

coupled_laplace_problem/coupled_laplace_problem.cc
DavidSCN: Deprecated

DavidSCN:
> Looks good to me. Thanks! I have left some minor comments. Once these are addressed, it is ready from my side.
> 
> Maybe someone else has some comments. I let someone else merge!

Thanks for the review! I addressed your comments and the code is ready.

peterrum (approved):
Thx!

jppelteret:
> Maybe someone else has some comments. I let someone else merge!

Hi @DavidSCN. Thanks a lot for putting this really interesting example together. Its a really exciting piece of work (and, to be honest, I could have made use of this functionality that preCICE offers 10 years ago :-D). I'm keen to give it a little look before we merge it, so I'll try my best to get that done on the weekend. I know that you, with the help of @uekerman and  @peterrum and maybe others, have spent a lot of time and effort to fine-tune what's here, so I'll try to keen my comments to be as concise and to the point as possible.

coupled_laplace_problem/CMakeLists.txt
```diff
@@ -0,0 +1,47 @@
...
+    "You may want to either pass a flag -DDEAL_II_DIR=/path/to/deal.II to cmake\n"
+    "or set an environment variable \"DEAL_II_DIR\" that contains this path."
+    )
+ENDIF()
+
+DEAL_II_INITIALIZE_CACHED_VARIABLES()
+PROJECT(${TARGET})
+DEAL_II_INVOKE_AUTOPILOT()
+
+FIND_PACKAGE(precice REQUIRED)
```
jppelteret, line 41: May we please add a `HINT` to where `preCICE` is located? We did the same for XBraid in [this example](https://github.com/dealii/code-gallery/blob/master/parallel_in_time/CMakeLists.txt#L61-L71), and that's really useful when the package is not specified in path.

coupled_laplace_problem/README.md
```diff
@@ -0,0 +1,69 @@
...
+
+* `deal.II`, version `9.2` or greater. Older versions might work as well, but have not been tested.
+
+* [preCICE](https://github.com/precice/precice/wiki#1-get-precice), version `2.0` or greater. Have a look at the provided link for an installation guide.
+
+## Compiling and running
+
+Similar to the example programs, run
+```
+cmake -DDEAL_II_DIR=/path/to/deal.II .
```
jppelteret, line 36: To accompany the CMake `HINT` that I suggested earlier, can we also document the `-DPRECICE_DIR` (or whatever you call the CMake variable) here?

coupled_laplace_problem/README.md
```diff
@@ -0,0 +1,69 @@
...
+```
+or
+```
+make release
+```
+This command will generate two executables: one for the `coupled_laplace_problem` and one for the `fancy_boundary_condition` participant.
+```
+./coupled_laplace_problem
+```
+excutes the `coupled_laplace_problem`. In order to start the coupled simulation, execute
```
jppelteret, line 51: `excutes` -> `executes`

coupled_laplace_problem/README.md
```diff
@@ -0,0 +1,69 @@
...
+```
+./fancy_boundary_condition
+```
+in the same directory from another terminal window.
+
+preCICE and the deal.II solver create during the simulation several log files. In order to remove all result-associated files and clean-up the simulation directory the Allclean script can be executed.
```
jppelteret, line 57: `create during the simulation several log files` -> `create several log files during the simulation`

coupled_laplace_problem/fancy_boundary_condition.cc
```diff
@@ -0,0 +1,117 @@
+// This program does not use any deal.II functionality and depends only on
+// preCICE and the standard libraries.
+#include <precice/SolverInterface.hpp>
+
+#include <iostream>
+#include <sstream>
+
+// The prgram computes a time-varying parabolic boundary condition, which is
```
jppelteret, line 8: `prgram` -> `program`

coupled_laplace_problem/fancy_boundary_condition.cc
```diff
@@ -0,0 +1,117 @@
...
+      }
+
+  // Pass the vertices to preCICE
+  precice.setMeshVertices(meshID,
+                          numberOfVertices,
+                          vertices.data(),
+                          vertexIDs.data());
+
+  // initialize the Solverinterface
+  double dt = precice.initialize();
```
jppelteret, line 90: Could this be made `const`?

coupled_laplace_problem/coupled_laplace_problem.cc
jppelteret: > Should I do it anyway?

No, it's OK thanks. You mentioned previously (in the code itself) that you closely followed step-4's original implementation, so the reader should be aware of this and could just adjust the implementation according to their own needs/expectations.

jppelteret (changes requested):
Thank you for your contribution @DavidSCN and @uekerman. This is really great, and I'm sure that preCICE is a library that many of our users would be very interested to know and learn more about. (Many years ago I had to implement a poor-man's version of what you're doing here to couple Fluent and my deal.II code, so I have a deep appreciation for the library is offering to us.) If you have any inclination to add more examples, you would have my support! I think that this one really gets to the heart of what preCICE can do, and would be a great building block for more sophisticated applications.

There were a couple of really small things (typos and a grammar error) that I've noted. Could I also ask you to make one small addition to the CMake script? It would make our lives a lot easier when we do a version bump on the code-gallery examples. Specifically, I would build and test the example with an out-of-path installation of preCICE so having a working hint to where the library is would help in that regard. I could also just do it in a follow up PR to this one -- I don't mind.

After that, perhaps you could squash all of your commits and I'd be happy to merge this awesome contribution.

coupled_laplace_problem/fancy_boundary_condition.cc
DavidSCN: No, it is potentially modified later

https://github.com/DavidSCN/code-gallery/blob/5cf7e24262d3eb35fd7c1905c5a6f8ce1dde3ed6/coupled_laplace_problem/fancy_boundary_condition.cc#L108

In this simple example, it is of course `const` and not modified later on. I would keep it for demonstration purposes.

coupled_laplace_problem/CMakeLists.txt
fsimonis: Hi! As this is using `find_package`, can use the build-in hints to pass this information to CMake.

Let's assume, you installed preCICE using the prefix `/opt/imism/precice/` then you may pass the location to CMake by either specifying
* `precice_DIR=/opt/imism/precice/lib/cmake`, or
*  `CMAKE_PREFIX_PATH=/opt/imism/precice`.

Both can be passed as CMake variables when calling `cmake` to configure the example project or by setting them as environment variables prior to calling `cmake`.
```
$ cd precice_tutorial/build

$ export precice_DIR=/opt/imism/precice/lib/cmake
$ cmake ..
or
$ cmake -Dprecice_DIR=/opt/imism/precice/lib/cmake ..
```

`CMAKE_PREFIX_PATH` is also used by spack to pass location information.

coupled_laplace_problem/fancy_boundary_condition.cc
jppelteret: Ah, of course. Silly of me to miss that.

jppelteret:
> I replaced now the find_package cmake tag by your proposed find_library tag and it works on my machine. In preCICE, we use the environment variable PRECICE_ROOT to specify the location.

Thank you for doing this!

> However, I'm not sure if the result is now what you were asking for. Should I keep the find_package module or should I remove it?

That XBraid example was a little more complicated due to how that package gets installed (the final directory structure is different if its hand installed vs installed via a package manager), so it wasn't the best example to direct you towards. But it is the only example in the code-gallery of linking to another package, so I chose it regardless... I understand why you thought that you needed to completely change the setup. Sorry that it wasn't clear what I was asking, but I meant that we could just have a hint added to the call to`find_package`.  Here's the [CMake documentation](https://cmake.org/cmake/help/latest/command/find_package.html#full-signature-and-config-mode) -- you can see that there's an optional parameter for the `HINT`.

coupled_laplace_problem/CMakeLists.txt
```diff
@@ -0,0 +1,75 @@
...
+  MESSAGE(FATAL_ERROR "\n"
+    "*** Could not locate a (sufficiently recent) version of deal.II. ***\n\n"
+    "You may want to either pass a flag -DDEAL_II_DIR=/path/to/deal.II to cmake\n"
+    "or set an environment variable \"DEAL_II_DIR\" that contains this path."
+    )
+ENDIF()
+
+IF(NOT "$ENV{PRECICE_ROOT}" STREQUAL "")
+	SET(PRECICE_ROOT "$ENV{PRECICE_ROOT}" CACHE INTERNAL "Copied PRECICE_ROOT from environment variable")
+ENDIF()
```
jppelteret, line 39: This code block can stay. With this and the hint in place then one can configure the project like this
```
cmake \
-DDEAL_II_DIR=$(spack location -i dealii@9.2.0) \
-DPRECICE_ROOT=$(spack location -i precice) \
-DCMAKE_BUILD_TYPE="Debug" \
..
```
and it should all work without any of these libraries that are installed (in this instance through Spack), being added to the environment paths.

jppelteret:
> Thanks for your review, great to hear that the contribution finds approval! 

You're welcome!

> I don't know how much you looked around, maybe you already noticed that we are actually using your structure codes with preCICE for FSI cases. The well-known FSI benchmarks are freshly validated :). Nevertheless, there is still some work open and -of course- if you run similar cases with preCICE, any contribution is welcome..

No, I had no idea that you used that code as a sample case and a component for benchmarking. That's really awesome, and it totally made my day to find out that it was being used in this way! Thank you for letting me know :-) Once I'm done with my current side projects (which may take some while, because I don't have much free time nowadays), I'd be keen to take a closer look at how you've hooked it up into that code and then I'll start to think about what I can do with it. I've been thinking about trying my hand at some FSI, but your framework give a totally different spin on how one can approach the problem...

DavidSCN:
> and it should all work without any of these libraries that are installed (in this instance through Spack), being added to the environment paths.

Ah ok, now I got it! As pointed out by @fsimonis (thanks!) there is already a build-in `HINT` named `precice_DIR` which can be used for this purpose. In order to avoid confusion regarding the small and capital spelling of precice, I added an additional `PRECICE_DIR` hint. Besides XBraid, an appropriate example is probably included in all deal.II related CMake scripts where the same instruction is used to locate deal.II.. Now the example can be used with `-DPRECICE_DIR` or with `-Dprecice_DIR` without a global installation of preCICE. I also added a status message in CMake in order to print the preCICE directory, added a respective note in the README file of this tutorial and finally squashed all commits together.

jppelteret (approved):

jppelteret:
Great, thanks a lot for adding that. And many thanks again for the great contribution! I'm sure that this will be of interest to many people.

> Besides XBraid, an appropriate example is probably included in all deal.II related CMake scripts where the same instruction is used to locate deal.II.

True ;-)

PR #90, opened 2021-07-26
Some fixes for functions of make_grid, make_constraints, and print_ve…

Lianxiali:
1. The index should be 2 for z coordinate in make_grid()
   
        else if (std::abs(std::abs(cell->face(face)->center()[**2**]) - 0.5) < tol_boundary)  // index 0 is changed to 2
           cell->face(face)->set_boundary_id(2); // +Z and -Z faces   
 
2. In make_constraints(), the conditional statement seems not to work as expected because if block and else block are the same.

           **two else blocks** are deleted.
 
3. In print_vertical_tip_displacement()， the following line triggers an error in case 3D and debug mode.

                  Point<dim> soln_pt (48.0*parameters.scale,60.0*parameters.scale);
     
           changed to:
                  Point<dim> soln_pt{}；
                  soln_pt[0] = 48.0*parameters.scale;
                  soln_pt[1] = 60.0*parameters.scale;


Quasi_static_Finite_strain_Compressible_Elasticity/cook_membrane.cc
```diff
@@ -1101,7 +1101,7 @@ namespace Cook_Membrane
               cell->face(face)->set_boundary_id(1); // -X faces
             else if (std::abs(cell->face(face)->center()[0] - 48.0) < tol_boundary)
               cell->face(face)->set_boundary_id(11); // +X faces
-            else if (std::abs(std::abs(cell->face(face)->center()[0]) - 0.5) < tol_boundary)
+            else if (std::abs(std::abs(cell->face(face)->center()[2]) - 0.5) < tol_boundary)
```
jppelteret, line 1104: We probably now need an `if (dim == 3)` check here.

Quasi_static_Finite_strain_Compressible_Elasticity/cook_membrane.cc
```diff
@@ -2109,12 +2105,6 @@ namespace Cook_Membrane
                                                    ZeroFunction<dim>(n_components),
                                                    constraints,
                                                    fe.component_mask(z_displacement));
-        else
-          VectorTools::interpolate_boundary_values(dof_handler_ref,
-                                                   boundary_id,
-                                                   ZeroFunction<dim>(n_components),
-                                                   constraints,
-                                                   fe.component_mask(z_displacement));
       }
```
jppelteret, line 2109: Can you please make the same modification that we did to step-44 (https://www.dealii.org/developer/doxygen/deal.II/step_44.html#Solidmake_constraints) and add the following block:
```
if (apply_dirichlet_bc)
  {
...
  }
else
  {
if (constraints.has_inhomogeneities())
  {
          AffineConstraints<double> homogeneous_constraints(constraints);
          for (unsigned int dof = 0; dof != dof_handler.n_dofs(); ++dof)
            if (homogeneous_constraints.is_inhomogeneously_constrained(dof))
              homogeneous_constraints.set_inhomogeneity(dof, 0.0);
          constraints.clear();
          constraints.copy_from(homogeneous_constraints);
        }
    }
```
This makes extensibility of this code more straight-forward, as inhomogeneous BC are then correctly dealt with in the non-linear solver.

jppelteret (commented):
Thank you for the fixes. I wonder how I made that error in the boundary ID assignment. Can you please verify that the solution remains correct in the 3d case?

Quasi_static_Finite_strain_Compressible_Elasticity/cook_membrane.cc
Lianxiali: A fix look like **else if (dim == 3 && std::abs(std::abs(cell->face(face)->center()[2]) - 0.5) < tol_boundary)** should work.

Lianxiali: 3d case looks correct but there does a slight difference from 2d case. I'm not sure if it is reasonable or the plane strain condition is not implemented correctly.

Quasi_static_Finite_strain_Compressible_Elasticity/cook_membrane.cc
Lianxiali: I got an violation of **det_F > NumberType(0.0)** with the else block handling inhomogeneous BC. 

It works if the if conditions is removed, i.e., executing **interpolate_boundary_values** at each iteration. But it is a waste and the attempt avoiding this fails.

jppelteret:
Hi @Lianxiali. Many thanks for the PR with these updates. I'm sure that they're fine as is, but I'd like to take a quick look at the effect on the solution to verify that there are no further hidden corrections to be made after these changes are accepted. I'll do my best to do this on the weekend. I appreciate your patience!

Quasi_static_Finite_strain_Compressible_Elasticity/cook_membrane.cc
jppelteret: Oh, that's probably because the [`constraints.clear();`](https://github.com/dealii/code-gallery/blob/master/Quasi_static_Finite_strain_Compressible_Elasticity/cook_membrane.cc#L2061) was left in place. This needed to be moved into the `if (apply_dirichlet_bc){}` condition.

Quasi_static_Finite_strain_Compressible_Elasticity/cook_membrane.cc
jppelteret: I'd expect there to be a slight difference, because the 2d case is neither plane strain nor plane stress, whereas the 3d case models plane strain. I would like to verify that we're still converging on the accepted solution in the 3d case, so I'll dig up the reference paper and double check this with your changes in place :-)

Lianxiali:
Don't accept the update. There are still some errors with the updates. Simply deleting the else block does not work. I expect your further test and correction then this code will be more robust.

jppelteret:
Hi @Lianxiali, I've pushed another commit that fixes the remaining build issues and implements some other corrections. Do you want to check them out and let me know if you agree with those changes?

Lianxiali:
> Hi @Lianxiali, I've pushed another commit to your branch that fixes the remaining build issues and implements some other corrections. Do you want to check them out and let me know if you agree with those changes?

Thank you for making the commit. I'll take a look at that later.

Lianxiali:
I checked the code. It runs well for both 2d and 3d, debug and release modes. Well done, thanks for your work, Jean-Paul! I totally agree to all your changes.

I'm new to Github and deal.II. Do I need to accept your commits or merge the branch to mater? Or I just make the full request directly with your commits?

jppelteret:
> I checked the code. It runs well for both 2d and 3d, debug and release modes. Well done, thanks for your work, Jean-Paul! I totally agree to all your changes.

Great, thank you too for your contribution! We're looking forward to the next one ;-)

> I'm new to Github and deal.II. Do I need to accept your commits or merge the branch to mater? Or I just make the full request directly with your commits?

We could have squashed the two commits into one, but it wasn't really necessary in this instance. So I merged the branch for us in the state that it was (you don't have merge privileges, so one of the developers needed to do that).

Lianxiali:
Great! I'll keep an eye on the updates of the code-gallery.

PR #96, opened 2022-02-02
Add a section on testing this program to the readme file.

bangerth:
The reviewers of our paper have asked us to provide alternative implementations of this program in matlab or python. To verify that our matlab implementation is correct, I've written a small test harness that takes known inputs and runs them through the program to produce outputs that can then be compared against what we compare in the reference implementation that this code gallery program represents.


bangerth:
I've got a follow-up that is going to step on this patch. Would someone be willing to take a look at it?

tjhei:
I am not strongly opposed to having these files in here, but I wonder if a separate repo for your publication might make more sense, especially if you are going to add matlab/python/whatever implementations or other data.

bangerth:
I think the code gallery offers a more permanent status than a random repo somewhere. It also has the advantage that the program remains updated.

The matlab/python scripts are both ~250 lines, so not very much.

bangerth:
What to do? I'd love to re-submit the paper this week, and would need to figure out what to do with this PR and with the addition of the matlab/python code.

My take is that we've generally left wide leeway to the authors of code gallery codes. There is no good reason from my perspective to have this code both in the code gallery and in another external repository -- that's just confusing. So if I had to put it into an external repo, I'd probably want to retract it from the code gallery. On a broader scale, I've always seen the code gallery as a place where authors could deposit the code used for their own papers; if that had included data sets that were used with that code in a paper, I would certainly have thought it appropriate to have those be part of the code gallery if they were reasonably sized. The input/output files here, along with alternative implementations, would fall into the same category in my book.

But that is my personal interpretation of what we want with the code gallery. I respect if others feel differently -- I just have to make a decision one way or the other this week.

masterleinad (approved):
I think it's reasonable to add these files.

tjhei:
I didn't mean to block this with my comment, sorry.

PR #102, opened 2022-06-28
Adjusted README.md and applied small 'coding design' modifications (using EvaluationFlags, modified template parameters...)

GiuseppeOrlando878776:


NavierStokes_TRBDF2_DG/README.md
```diff
@@ -26,25 +26,25 @@ In this section, we briefly describe the problem and the approach employed. A de
...
+\nabla\cdot\mathbf{u} &= 0,
 \end{align*}
 $$
 
-where $Re$ denotes the Reynolds number. In the case of projection methods, difficulties arise in choosing the boundary conditions to be imposed for the Poisson equation which is to be solved at each time step to compute the pressure. An alternative that allows to avoid or reduce some of these problems is the so-called artificial compressibility formulation. In this formulation, the incompressibility constraint is relaxed and a time evolution equation for the pressure is introduced, which is characterized by an artificial sound speed $ c, $ so as to obtain:
+where $Re$ denotes the Reynolds number. In the case of projection methods, difficulties arise in choosing the boundary conditions to be imposed for the Poisson equation which is to be solved at each time step to compute the pressure. An alternative that allows to avoid or reduce some of these problems is the so-called artificial compressibility formulation. In this formulation, the incompressibility constraint is relaxed and a time evolution equation for the pressure is introduced, which is characterized by an artificial sound speed $c,$ so as to obtain:
```
marcfehling, line 33:
```suggestion
where $Re$ denotes the Reynolds number. In the case of projection methods, difficulties arise in choosing the boundary conditions to be imposed for the Poisson equation which is to be solved at each time step to compute the pressure. An alternative that allows to avoid or reduce some of these problems is the so-called artificial compressibility formulation. In this formulation, the incompressibility constraint is relaxed and a time evolution equation for the pressure is introduced, which is characterized by an artificial sound speed $c$, so as to obtain:
```

NavierStokes_TRBDF2_DG/README.md
```diff
@@ -61,15 +61,18 @@ $$
 Notice that, in order to avoid solving a nonlinear system at each time step, an approximation is introduced in the nonlinear momentum advection term, so that $\mathbf{u}^{n + \frac{\gamma}{2}}$ is defined by extrapolation as
 
 $$
+\begin{align*}
 \mathbf{u}^{n + \frac{\gamma}{2}} = \left(1 + \frac{\gamma}{2\left(1-\gamma\right)}\right)\mathbf{u}^{n} - \frac{\gamma}{2\left(1-\gamma\right)}\mathbf{u}^{n-1}.
+\end{align*}
 $$
```
bangerth, line 67: I'm not sure how this is going to look like, but let's see :-)

PR #107, opened 2022-07-18
Fix some code gallery programs.

bangerth:
This is not enough to make all of them compile, much less compile without warning. But it is a step in the right direction and I need to go home. I would appreciate a review, and I'll get to the other issues at a later time (tomorrow?).

### DIFF
--- a/Distributed_LDG_Method/LDGPoisson.cc
+++ b/Distributed_LDG_Method/LDGPoisson.cc
@@ -330,8 +330,8 @@ make_dofs()
   DoFTools::extract_locally_relevant_dofs(dof_handler,
                                           locally_relevant_dofs);
 
-  std::vector<types::global_dof_index> dofs_per_component(dim+1);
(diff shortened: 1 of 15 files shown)


time_dependent_navier_stokes/time_dependent_navier_stokes.cc
```diff
@@ -698,7 +698,7 @@ namespace fluid
         VectorTools::interpolate_boundary_values(
           dof_handler,
           id,
-          Functions::ZeroFunction<dim>(dim + 1),
+          Functions::Functions::ZeroFunction<dim>(dim + 1),
```
Rombur, line 701: I don't think this is correct.

Rombur (changes requested):

bangerth:
I will bump the required version in a separate PR.

Rombur (approved):

marcfehling (approved):

PR #110, opened 2022-07-19
Fix the remaining compiler errors.

bangerth:
I live under no illusion that everything will just work as expected, but being able to compile is a good first step :-)

### DIFF
--- a/Nonlinear_PoroViscoelasticity/nonlinear-poro-viscoelasticity.cc
+++ b/Nonlinear_PoroViscoelasticity/nonlinear-poro-viscoelasticity.cc
@@ -49,7 +49,6 @@
 #include <deal.II/grid/grid_out.h>
 #include <deal.II/grid/manifold_lib.h>
 #include <deal.II/grid/tria_accessor.h>
-#include <deal.II/grid/tria_boundary_lib.h>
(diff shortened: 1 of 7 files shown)


Nonlinear_PoroViscoelasticity/nonlinear-poro-viscoelasticity.cc
```diff
@@ -4005,7 +4002,7 @@ namespace NonLinearPoroViscoElasticity
                                      0.5);
 
             const double rot_angle = 3.0*numbers::PI/2.0;
-            GridTools::rotate( rot_angle, 1, this->triangulation);
+            GridTools::rotate( Point<3>::unit_vector(1), rot_angle, this->triangulation);
```
marcfehling, line 4005: I didn't know this function existed! It returns a `Point` object, but it will be implicitly converted to a `Tensor<1>` for the `rotate` function. Okay with me :)

marcfehling (approved):

Nonlinear_PoroViscoelasticity/nonlinear-poro-viscoelasticity.cc
bangerth: Yes, correct.

PR #111, opened 2022-07-19
Improve mark-up.

bangerth:
The alignment char inserts awkward spacing.


marcfehling (commented):
Hmm, I guess the intention of the authors was to actually use the features of the `align` environment, but simply forgot to put the `&=` instead of the `=` in the second row of each block.

Well, it works both ways :)

PR #116, opened 2022-07-25
Update cdr

drwells:
I cleaned things up a little while investigating the ML problem.

### DIFF
--- a/cdr/readme.md
+++ b/cdr/readme.md
@@ -1,43 +1,30 @@
 ## Overview
 I started this project with the intent of better understanding adaptive mesh
-refinement, parallel computing, and <code>CMake</code>. In particular, I started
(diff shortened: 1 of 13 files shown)


cdr/readme.md
```diff
@@ -1,43 +1,34 @@
...
+  and produce output.
 
 ## Requirements
-<ul>
-  <li>A <code>C++-11</code> compliant compiler</li>
-  <li>Version 8.4 of <code>deal.II</code></li>
-</ul>
-
+* A `C++-11` compliant compiler
+* Version 8.4 of `deal.II`
```
bangerth, line 31: What do you think about just removing this? Both are prereqs already upstream.

drwells: Good idea.

PR #118, opened 2022-09-02
Update a couple of numbers.

bangerth:
One of the previous PRs for this code gallery program accelerated computations by about a factor of 4, but didn't update two comments in the program itself. There is also a comment in the readme file that was never actually correct :-)


tjhei (approved):
I was wondering about these numbers when I played with the program last week. I was just thinking my laptop is faster... :-)

PR #117, opened 2022-08-14
Add program for goal-oriented Maxwell eigenvalue hp-refinement

harmonj:
Adds a code gallery project for goal-oriented hp-refinement for eigenvalue quantities of interest of the Maxwell eigenvalue problem for a 2-D benchmark that includes eigenpairs with singular eigenfunctions.

This is tested and working with the latest version of deal.II (9.4).

PR #120, opened 2023-03-11
Improved default params; Readme edit

harmonj:
Follow up to #117. Adjusts default parameters such that debug mode execution time is not unbearable.

### DIFF
--- a/Maxwell-Eigenvalue-hp-Refinement/Readme.md
+++ b/Maxwell-Eigenvalue-hp-Refinement/Readme.md
@@ -25,10 +25,10 @@ $$ e_{\lambda_{hp}} := \lambda-\lambda_{hp}. $$
 Some comments on the discretization of Maxwell's equations
 ----------------------------------------------------------
 In the proper solution of variational problem, $V_{hp}$ is not arbitrary, but should instead a subspace of $H(\mathrm{curl};\,\Omega)$ or, with the boundary conditions indicated above, $H_{0}(\mathrm{curl};\,\Omega)$, where
-$$ H(\mathrm{curl};\,\Omega) = \left\{\textbf{u}\in \left[ L_2(\Omega)\right]^d \, \mathrm{s.t.} \, \nabla\times\textbf{u}\in \left[ L_2(\Omega)\right]^{d} \right\}$$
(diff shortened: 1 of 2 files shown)


Maxwell-Eigenvalue-hp-Refinement/Readme.md
```diff
@@ -25,10 +25,10 @@ $$ e_{\lambda_{hp}} := \lambda-\lambda_{hp}. $$
 Some comments on the discretization of Maxwell's equations
 ----------------------------------------------------------
 In the proper solution of variational problem, $V_{hp}$ is not arbitrary, but should instead a subspace of $H(\mathrm{curl};\,\Omega)$ or, with the boundary conditions indicated above, $H_{0}(\mathrm{curl};\,\Omega)$, where
-$$ H(\mathrm{curl};\,\Omega) = \left\{\textbf{u}\in \left[ L_2(\Omega)\right]^d \, \mathrm{s.t.} \, \nabla\times\textbf{u}\in \left[ L_2(\Omega)\right]^{d} \right\}$$
+$$ H(\mathrm{curl};\,\Omega) = \left\{\textbf{u}\in \left[ L_2(\Omega)\right]^d \, \mathrm{s.t.} \, \nabla\times\textbf{u}\in \left[ L_2(\Omega)\right]^{2d-3} \right\}$$
```
bangerth, line 28: Oh, my apologies. I had assumed that that was a typo. I do not understand what this is supposed to mean, then, though. H(curl) is the space for which the function is in L2 and its curl is in 2d. Are you using the notation here to suggest that in 2d, the curl is scalar and in 3d it is a vector?

bangerth (approved):
Assuming my comment is correct, I'll merge.

Maxwell-Eigenvalue-hp-Refinement/Readme.md
harmonj: Yes, this definition is consistent with the literature. If we take our domain as 2D with transversal fields, the curl is necessarily axial, which is most precisely described in this definition.

bangerth: I see. I had never seen anyone use this notation.

PR #136, opened 2023-04-14
Restrict permissions in CI.

marcfehling:
Allows to manually trigger a github workflow. Useful in combination with the scheduler, when a critical change in deal.II should be tested immediately.

Also restrict permissions for security reasons.

### DIFF
--- a/.github/workflows/precice-ci.yml
+++ b/.github/workflows/precice-ci.yml
@@ -8,6 +8,10 @@ on:
   schedule:
   - cron: '0 0 * * 0'
 
+permissions:
+  contents: read
+  issues: write
+
 jobs:
   build:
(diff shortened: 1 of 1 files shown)


marcfehling:
I removed the [`workflow_dispatch`](https://github.com/dealii/code-gallery/pull/136/commits/be6180ab1d3ac70aed02c49988287f56c3bae374) option again for security reasons. There is no clarity who can start the workflow.

I'll leave a link to an ongoing discussion here. We can add the feature once there is a solution and in case we really need it.
https://github.com/orgs/community/discussions/49688

marcfehling:
This patch closes a security issue and is only 4 lines long. Please consider merging it soon :)

masterleinad (approved):

PR #137, opened 2023-04-15
Test jammy releases.

marcfehling:
Follow-up to https://github.com/dealii/dealii/pull/15074.

Switch from focal to jammy images and releases.

We could also test both focal and jammy, but I don't see much benefit in it. Opinions?


marcfehling:
Rebased on master.

drwells (approved):
One should be fine - we'll know if things break due to the other change.

PR #141, opened 2023-04-17
Automatically create issue when scheduled CI is failing.

marcfehling:
As mentioned in https://github.com/dealii/dealii/issues/15073#issuecomment-1502410198. I followed [this guide](https://epiforecasts.io/posts/2022-04-11-robust-actions/#notify-the-whole-team-when-a-scheduled-workflow-fails) essentially.

### DIFF
--- a/.github/action-issue-template.md
+++ b/.github/action-issue-template.md
@@ -1,0 +1,5 @@
+---
+title: "Scheduled GitHub Actions worker is failing"
+---
+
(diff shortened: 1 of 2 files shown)


drwells (approved):
Nice! I suppose we won't be sure if it works or not until this breaks again.

marcfehling:
> Nice! I suppose we won't be sure if it works or not until this breaks again.

I played around with that feature on my fork. I chose the settings specifically to match the dynamics in our group.

We won't be notified when a scheduled worker fails -- this neat feature gives us more visibility.

PR #142, opened 2023-04-23
CI now covers compilation of all projects.

marcfehling:
Part of https://github.com/dealii/dealii/issues/15073.

I believe this is the most convenient and low-maintenance way to verify that all code-gallery programs compile.

`CeresFE` couldn't be compiled with deal.II 9.4 because of a deprecated function that has been deleted. I've updated it in this patch so that the CI will pass.

`parallel_in_time` finds the braid include folder, but has issues finding the actual library. I've changed a few lines in its `CMakeLists.txt`.

### DIFF
--- a/parallel_in_time/CMakeLists.txt
+++ b/parallel_in_time/CMakeLists.txt
@@ -48,24 +48,16 @@ IF(NOT ${deal.II_FOUND})
 ENDIF()
 
 # Find braid details
-IF(NOT "$ENV{BRAID_DIR}" STREQUAL "")
-    SET(BRAID_DIR "$ENV{BRAID_DIR}" CACHE INTERNAL "Copied BRAID_DIR from environment variable")
-ENDIF()
-
-IF("${CMAKE_FIND_LIBRARY_PREFIXES}" STREQUAL "")
-    SET(CMAKE_FIND_LIBRARY_PREFIXES "lib")
-ENDIF()
+SET(BRAID_DIR "$ENV{BRAID_DIR}" CACHE INTERNAL "Copied BRAID_DIR from environment variable")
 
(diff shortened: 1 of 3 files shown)


marcfehling:
In later patches, we might want to
-  be more rigorous about warnings with `-Wall -Werror`
-  check the output of each individual program similar to `coupled_laplace_problem`

parallel_in_time/CMakeLists.txt
```diff
@@ -48,24 +48,16 @@ IF(NOT ${deal.II_FOUND})
...
   NAMES braid.h
   HINTS ${BRAID_DIR}
-  PATH_SUFFIXES include
   )
 
 FIND_LIBRARY(BRAID_LIBRARY
-  NAMES braid.a
+  NAMES libbraid.a
   HINTS ${BRAID_DIR}
-  PATH_SUFFIXES lib64 lib
```
masterleinad, line 68: Normally, you would install `xbraid` instead of just building it, right? I think we should do that in the CI, too.

bangerth: xbraid is such a homegrown project that it is built with a small `Makefile` and no `install` target :-)

masterleinad: Ah, OK.

marcfehling: There is no convenient way to install it using their Makefile. Building is also the option they provide in their user manual.
https://github.com/XBraid/xbraid/files/10865642/user_manual.pdf
https://github.com/XBraid/xbraid/blob/master/braid/Makefile

masterleinad (approved):

PR #145, opened 2023-05-23
parallel in time: Convert tabular to markdown.

marcfehling:
Doxygen doesn't like tables as LaTeX environments.
https://www.dealii.org/developer/doxygen/deal.II/code_gallery_parallel_in_time.html
image

The doxygen documentation suggests to use markdown tables.
https://www.doxygen.nl/manual/tables.html

This PR converts the tabular environment in the parallel in time code to a markdown table.


masterleinad (approved):

marcfehling:
Looks like this now:
image

PR #146, opened 2023-05-26
Swift hohenberg solver

Sam0h51:
This piece of code solves the generalized Swift-Hohenberg equation on many of the 2 dimensional manifolds built into dealii. It's set up to allow for testing across a wide range of parameters, which can produce interesting differences in the patterns.


bangerth:
I had looked at this before and made some comments, so will take the liberty to merge as-is.

I will note that @Sam0h51 is one of the students in this year's deal.II class here at CSU.

bangerth:
This is now online: https://dealii.org/developer/doxygen/deal.II/code_gallery_Swift_Hohenberg-Solver.html

I'm on vacation, and so won't investigate why this is so, but the pictures are all not correctly shown inlined into the text. We need to address this before the release.

Sam0h51:
Hi Wolfgang,

Do you know what version of markdown the dealII  webpage uses? I ran into
issues where the README looked correct in visual studio but not on GitHub,
so I’d assume it’s a similar issue here. I should be able to fix it from
there.

Hope your vacation is going well!

Best,
Sam

PR #149, opened 2023-06-20
Fix formulas in the Swift-Hohenberg program

bangerth:
Formulas are currently not correctly shown in this program. This patch tries to fix this. While there, also convert from DOS to Unix line endings.


bangerth:
I would like to get this merged sooner rather than later so that we can still see whether this actually worked in time for generating the release documentation.

marcfehling (approved):

bangerth:
OK, this now looks good: https://dealii.org/developer/doxygen/deal.II/code_gallery_Swift_Hohenberg-Solver.html

PR #157, opened 2023-08-07
Fix quotation issue.

bangerth:
I *think* that the formatting issue on https://dealii.org/current/doxygen/deal.II/code_gallery_Nonlinear_PoroViscoelasticity.html (reported on the mailing list) is caused by the backtick quoting in this file. Let's just use regular quotes.


marcfehling:
Looks good again!
https://www.dealii.org/developer/doxygen/deal.II/code_gallery_Nonlinear_PoroViscoelasticity.html

PR #158, opened 2023-08-08
Update the Maxwell-hp example.

### LINKED ISSUE #153: maxwell-hp no longer compiles.
See [the action log](https://github.com/dealii/code-gallery/actions/runs/5773659504) for more details.

drwells:
Fixes #153.

We made some breaking changes right after the 9.5 release. Lets keep things working both with the most recent release and the present prerelease.


marcfehling:
The `maxwell-hp` program works again. There is now a different issue in `NS_TRBDF2_DG` which we need to address separately.

marcfehling (approved):

PR #160, opened 2023-08-09
Update the NavierStokes TRBDF2 DG example - part 2.

marcfehling:
#159 updated the `gather_evaluate` functions, but I missed that `evaluate` and `integrate` need a similar update.
https://www.dealii.org/9.5.0/doxygen/deal.II/classFEEvaluation.html#a706df1ddc8e95119188be47a04e5625e
https://www.dealii.org/9.5.0/doxygen/deal.II/classFEEvaluation.html#a311c0ac5cf4b8cf2458996d7099a139a


marcfehling:
Would some give this a quick look and merge? Then the worker scheduled for this weekend will finally pass again :)

PR #161, opened 2023-09-08
Do not use deprecated extract_locally_* functions in code-gallery.

marcfehling:
Follow-up to https://github.com/dealii/dealii/pull/15966.


bangerth (approved):
Thanks!

PR #167, opened 2024-02-03
Update precice to v3.

marcfehling:
Let's see if `coupled_laplace_problem` is happy.

PR #168, opened 2024-02-06
Port preCICE example to v3

davidscn:
preCICE version 3 comes with a couple of breaking changes. This PR applies all changes necessary. Our [porting guide](https://precice.org/couple-your-code-porting-v2-3.html) still needs some clean-up, but should help with such things in the near future.

With the current changes, everything works technically and I would expect the same results as obtained with the v2.5.0 (CI ref data). However, 

- [ ] some code comments still need an update 
- [ ] for exemplary purposes, I would like re-arrange the order of some API calls. It shouldn't make a difference with this trivial example case, but might lead to misunderstandings for realistic examples. See also the detailed description in [our doxygen docs](https://precice.org/doxygen/develop/classprecice_1_1Participant.html#Detailed%20Description)
- [ ] Not sure about the formatting changes here. I only have clang-format 15 as of now (assuming the clang-format is just the same as the main deal.II project)
- [ ]


marcfehling:
I'm sure users who would like to use precice and deal.II will at some point stumble over your code gallery program. So please adjust it in all means you find helpful for understanding the code. I second your suggestions on updating comments and changing the order of API calls.

Regarding indentation: That choice is entirely up to you. I like the fact that you would like to adhere to a certain indentation style. In deal.II we currently use clang-format 16 (see [here](https://github.com/dealii/dealii/blob/e61fb518c39ad06d19b8913451c72151b24b80e2/contrib/utilities/download_clang_format#L27)), but it's fine if you use the tools you have available.

Also, feel free to take over/cherry-pick the commit from #167 to have a coherent patch.

davidscn:
Thanks for the reply. Changes went now a bit bigger than anticipated, as the handling of time and time-related data becomes more explicit. I changed the `fancy_boundary_condition` to actually provide data related to `t=0.1` and made the data of `t=0` optional (it's only used in case `initialize="true"` is used in the preCICE configuration file, which is not the case by default). With this, the association of time and the boundary condition is actually correct and also the relevant data for the Laplace participant, as it uses a backward Euler scheme, i.e., coupling data is required at the end of each time window. 

Due to the changes time management, I had to update the reference/CI data as well. I verified, that all technical changes lead to the same results obtained with v2.5.0.

Ready for review.

marcfehling (approved):

marcfehling:
Thank you for the updates!

PR #170, opened 2024-02-23
Update license information

tamiko:
Here is a preview: https://github.com/tamiko/code-gallery/blob/update_license_information/LICENSE.md


tamiko:
All I suggest the following license information with a relatively permissive license policy.

bangerth:
Can we do this for the code gallery? Do we want to do this?

tamiko:
@bangerth
> Can we do this for the code gallery?

Yes.

Let me give some context: My suggested rules only affect new contribution to the code gallery. In essence we now require for every contribution
 - to be covered by the "Developer Certificate of Origin" (https://developercertificate.org/). This document was developed by the Linux Foundation and is used for code contributions to the linux kernel. In short this document requires a contributor to certify that they are actually entitled to contribute (e.g., either hold the copyright, or have the OK from all copyright holders). We have used this for the main deal.II repository for many years.
 - The second statement requires that the code is actually released under an open source license. I mean what is the point of adding something to the code gallery if no one can use it?

> Do we want to do this?

There are many ways moving forward including not changing anything. I have opened two pull requests that do the following:
 - We clarify in the `LICENSE.md` that we spell out the authors and licenses that apply in individual source code.
 - In #171 I have changed all contributed code to have a proper header that lists the author and the currently applicable license, which I assumed is `LGPL-2.1-or-later` because that is what we had in the `LICENSE` file. But we never required individual authors to affix a header, or make an explicit statement.

I think this merely spells out the current status quo.

We could also take the stance that the same license should apply that we have in our main repository.

luca-heltai (approved):

gassmoeller (approved):
I think this makes a lot of sense. Having the code gallery use the same licenses as the main library avoids confusion, and as you say this PR only applies to newly contributed code. All existing examples will be listed as LGPL-2.1 or later, all new examples will be dual licensed.

I will wait a few days for @bangerth to give his ok though before merging.

bangerth (approved):

PR #171, opened 2024-02-23
Update copyright headers

tamiko:


tjhei (approved):
Makes sense to me. Would it make sense to add email address(es) of the main authors in the headers as well? Or too much effort? I think it makes sense to add them in the future at least to make it easier for users to get in touch.

tamiko:
@tjhei We record all contributors with e-mail in `doc/author` so I figured just the name in the copyright header is enough.

bangerth (approved):

PR #173, opened 2024-02-24
switch to unix line endings, add header

tamiko:


gassmoeller (approved):
Is this the same as #171 except only the files with windows line endings?

Makes sense to me, depending on the resolution of the conversation in #171 the same changes should be made here (or not), but otherwise this is good to go.

tamiko:
@gassmoeller Thanks for reviewing. Yes, I split this off into a separate PR due to the amount of line changes.

bangerth (approved):

PR #180, opened 2024-06-24
Fix warnings.

marcfehling:
Fixes some unused-parameters/variables/functions and suggest-override warnings.


nonlinear-heat_transfer_with_AD_NOX/source/boundary_values.cc
```diff
@@ -11,14 +11,16 @@
  */
 double Boundary_values_left::value(const Point<2> &p, const unsigned int) const
```
masterleinad, line 12:
```suggestion
double Boundary_values_left::value(const Point<2> & /*p*/, const unsigned int) const
```

masterleinad (approved):

PR #179, opened 2024-06-23
Add the 'nonlinear heat transfer with AD/NOX' code gallery program.

bangerth:

PR #178, opened 2024-06-18
Nonlinear heat transfer problem

narasimhan-2604:
A simple nonlinear heat transfer problem is solved with automatic differentiation. The nonlinear equations are solved using TRILINOS NOX .

PR #181, opened 2024-06-24
Convert latex documentation to doxygen input.

bangerth:
This is a follow-up to #178 and #179 that takes @narasimhan-2604 's latex source files and converts them into doxygen input so that the documentation is shown as part of the README file.


bangerth:
Ping? There is no new material here, I just transcribed @narasimhan-2604 original latex sources into doxygen input.

marcfehling (approved):

PR #189, opened 2024-07-10
Do not use mark-up in section headings.

bangerth:
Apparently this does not work, we only get to see the original text.


bangerth:
See here: https://dealii.org/developer/doxygen/deal.II/code_gallery_TravelingWaves.html

magomedovs:
Indeed, Professor. Thank you for the corrections!

kronbichler (approved):

marcfehling (approved):

PR #204, opened 2024-10-23
Update linux workflow.

marcfehling:


masterleinad:
What's the motivation?

marcfehling:
I was looking for literature about deal.II and applications in elasticity, found the [deal.II-preCICE](https://github.com/precice/dealii-adapter) adapter, and noticed that precice received an update.

bangerth (approved):

PR #205, opened 2024-11-02
Fix formula markup in one code gallery program.

bangerth:
The formulas in https://dealii.org/developer/doxygen/deal.II/code_gallery_Maxwell_Eigenvalue-hp-Refinement.html are all mangled because our script does not understand `$$...$$` formulas. This patch fixes that.

@harmonj FYI.


bangerth:
Ping?

Rombur (approved):

PR #206, opened 2024-11-26
Try getting formulas right.

bangerth:
For reasons not clear to me, #205 did not actually fix things. https://dealii.org/developer/doxygen/deal.II/code_gallery_Maxwell_Eigenvalue-hp-Refinement.html still shows the broken formulas -- in fact, the page is then simply cut off.

I don't know what causes this, because with the same doxygen version (1.9.8) on my local system I get the complete picture, with formulas showing correctly. So let's try something.


bangerth:
This still does not seem to work. Does anyone have an idea what might be going on here?

PR #233, opened 2025-11-24
Agglomeration example

fdrmrc:
This PR presents a DG solver for the Poisson problem on general polytopal meshes generated through mesh agglomeration. Starting from a fine mesh, we construct a coarser agglomerated mesh on which the DG discretization is performed.

This set of files have been extracted from the [Polydeal](https://github.com/fdrmrc/Polydeal) library, a deal.II-based framework for working with agglomerated grids.

While the overall structure of `poisson.cc` largely follows that of a standard Poisson solver in deal.II, differences related to the usage of agglomerated meshes still need to be commented, so the PR remains a draft for now.


agglomeration_poisson/CMakeLists.txt
```diff
@@ -0,0 +1,47 @@
...
+  )
+
+# Add include directory
+INCLUDE_DIRECTORIES(include)
+
+# Usually, you will not need to modify anything beyond this point...
+
+CMAKE_MINIMUM_REQUIRED(VERSION 3.13.4)
+
+FIND_PACKAGE(deal.II 9.7.0
```
marcfehling: Thank you for your contribution!

Can you lower the requirement to deal.II 9.5.0? We at some point established that code gallery programs should compile with 9.5.0 and younger.

If your code does not compile with the older versions, you can work around it like in https://github.com/dealii/code-gallery/commit/42e8901f6b7954ecb043cce7f8a54c6f57acbda7.

bangerth: @marcfehling Do we need this requirement? Could we allow programs to just require newer versions (or in fact change *all* programs to just build for, say, 9.6)?

bangerth (approved):
Thank you, @fdrmrc and collaborators, and my apologies for letting this sit for so long. The code gallery is supposed to be a place where we don't impose exacting standards, and we should be merging programs quicker!

bangerth (commented):
Nice program! I did not look through everything, but I think it's a cool example of what one can do!

I really have only one comment, namely whether we could perhaps take the coverage stuff out of the cmake file? It is specific to GCC and/or clang, and will not work for the Windows compiler. I would like to see whether we can keep these programs available for all platforms. If you agree, either remove these lines or tell me and I can do it myself!

bangerth (approved):
Let's do it then!

bangerth:
The program is now here: https://dealii.org/developer/doxygen/deal.II/code_gallery_agglomeration_poisson.html Most of it looks good, but the images are not shown inlined into the readme text. Let me know if you'd like me to look into it, or if you see right away what the issue is.

xinpinggui:
Sorry for the late reply, and thank you for pointing this out.

I have now looked into the issue. The inline images were written using HTML tags such as

``

which render well on GitHub, but do not seem to be displayed properly on the deal.II code gallery page. I have replaced them with standard Markdown image syntax, for example

`image`.

I also noticed a few related formatting issues in the code gallery entry, including the rendering of some comments in `agglomeration_poisson.cc` and the layout of the references section. I will now open a follow-up pull request to fix these documentation and formatting issues.

xinpinggui:
Thanks again for pointing this out. I have now opened the follow-up pull request to address these image display and related formatting issues: PR #244.

PR #242, opened 2026-04-23
Keep GitHub actions alive

masterleinad:
The Actions didn't run on recent pull requests since GitHub disables them after 60 days of inactivity.


marcfehling (approved):
That is neat. Thank you! I will add it to the docker repositorty as well.

PR #243, opened 2026-04-24
Avoid the use of deprecated functionality.

bangerth:
Addresses some of the warnings we get in #242.


tjhei (approved):

tjhei:
with this you are breaking all 9.6 builds...

bangerth:
Oh, that's awkward. Let's not do that. I had imagined that the warnings we get are for 9.6 already, but that's clearly not the case. I'll close this, we can revisit the patch once we require 9.7.

PR #237, opened 2026-01-21
Add the code for ALE large-deformation thermoplasticity

maieneuro:
Following [this thread](https://groups.google.com/g/dealii/c/9AsfK9dZg5Q/m/2FqaJkKtBwAJ) in the mailing list, I'm adding this to the code gallery.

PR #252, opened 2026-07-03
ALE_Finite_Strain_Plasticity: fix deal.II 9.7 assertion, build cleanly, and run under MPI

maieneuro:
## Summary

Fixes the assertion reported in #237 and makes the
`ALE_Finite_Strain_Plasticity` code-gallery example build warning-free and
run correctly (serial **and** parallel) against deal.II 9.7.

The example previously aborted immediately in Debug, and (once that was
resolved) hit two further latent, parallel-only failures. All are fixed
here across two commits.

## Correctness fixes

**1. `ComponentMask` size mismatch (the assertion from #237)**
`mech_fe` was declared as `FESystem(FE_Q, dim, FE_Q, 2)`, giving `dim+2`
components, while the rest of the program consistently assumes `dim+1`
(displacement + angular twist): the `ComponentMask`s,
`LBCSystem<dim, Number, dim+1>`, the `zero_function`, `Material<dim+1>`,
the `FEValuesExtractors` (displacements `0..dim-1`, angular velocity at
`dim`) and the output component names/interpretations. The extra
component made

    component_mask.represents_n_components(fe.n_components())

fail inside `VectorTools::interpolate_boundary_values()`. Changed the
multiplicity `2 -> 1` so the FE has `dim+1` components, matching
everything else.

**2. Singular `MixedFEProjector` over `mech_fe`**
Removed a dead projector setup whose result (`mech_fe_projectors`) was
never used. Building an L2 projector over the full `mech_fe` (12 dofs/cell
for Q1) from only 4 quadrature points yields a rank-deficient mass matrix,
which aborted in `FullMatrix::gauss_jordan()` once the `ComponentMask`
assertion was past.

**3. `compress()` on ghosted vectors (parallel abort)**
`previous_deformation`, `previous_time_derivative` and `current_increment`
carry ghost elements (reinit with `locally_relevant_dofs`) and are
read-only. The code assigned a locally-owned vector into them and then
called `compress()`, which 9.7 forbids
(`has_ghost_elements() == false`). Assigning owned -> ghosted already does
the ghost import, so these `compress()` calls were both unnecessary and
illegal, matching the existing pattern in
`NewtonStepSystem::advance_time()`. Removed the 11 offending calls; kept
`compress()` on locally-owned vectors.

**4. Out-of-bounds MPI buffer access in `remap_material_state_variables`
(parallel abort)**
The point-to-point exchange passed buffer addresses as
`&group.at(i)[0]`. When a rank exchanges no points with rank `i` that
buffer is empty, and `operator[](0)` on an empty `std::vector` aborts
under `_GLIBCXX_ASSERTIONS` (enabled in deal.II Debug). Replaced all 12
`&..._groups.at(i)[0]` buffer arguments with `.data()`, which is well
defined for empty vectors.

## Warning cleanup

Debug and Release now compile with **zero project warnings**:
- `-Wdeprecated-declarations`: use the returning form of
  `extract_locally_relevant_dofs()` and the two-index-set
  `AffineConstraints::reinit()`.
- `-Wsuggest-override`: added `override` to the `Material` /
  `Function` / `std::exception` overrides.
- `-Wreorder`: fixed member-init order in
  `BoundaryUnidirectionalPenaltySpec` and `ScaleComponentFunction`.
- `-Wsign-compare`: cast `boundary_id()` comparisons to
  `types::boundary_id`.
- `-Wcatch-value`: catch `MaterialDomainException` by const reference.
- `-Wunused-*`: tagged genuinely-unused locals/parameters
  `[[maybe_unused]]` (preserving computations), dropped an unused local,
  and elided unused parameter names in stub methods.

## Testing

- **Debug** and **Release** build with 0 errors / 0 project warnings
  (deal.II 9.7).
- `mpirun -n 4` and `mpirun -n 18` run the full nonlinear solve
  (mesh-motion -> mechanical Newton iterations -> contact) without aborting.

## Commits

- `ALE_Finite_Strain_Plasticity: fix ComponentMask assertion and clean up build`
- `ALE_Finite_Strain_Plasticity: fix parallel (MPI) runtime aborts`

### DIFF
--- a/ALE_Finite_Strain_Plasticity/src/BoundaryUnidirectionalPenaltySpec.h
+++ b/ALE_Finite_Strain_Plasticity/src/BoundaryUnidirectionalPenaltySpec.h
@@ -30,9 +30,9 @@ namespace PlasticityLab {
 
   private:
     const unsigned int boundary_id;
-    const Number quadratic_spring_factor;
     const Number reference_displacement_increment;
     const Number residual_force;
+    const Number quadratic_spring_factor;
   };
 
 } /* namespace PlasticityLab */
--- a/ALE_Finite_Strain_Plasticity/src/DoFSystem.h
+++ b/ALE_Finite_Strain_Plasticity/src/DoFSystem.h
@@ -50,10 +50,9 @@ namespace PlasticityLab {
   void DoFSystem<dim, Number>::setup_dof_system (const FiniteElement<dim> &fe) {
     dof_handler.distribute_dofs(fe);
     locally_owned_dofs = dof_handler.locally_owned_dofs();
-    locally_relevant_dofs.clear();
-    DoFTools::extract_locally_relevant_dofs(dof_handler, locally_relevant_dofs);
+    locally_relevant_dofs = DoFTools::extract_locally_relevant_dofs(dof_handler);
 
-    nodal_constraints.reinit(locally_relevant_dofs);
+    nodal_constraints.reinit(locally_owned_dofs, locally_relevant_dofs);
     DoFTools::make_hanging_node_constraints (dof_handler, nodal_constraints);
   }
 
--- a/ALE_Finite_Strain_Plasticity/src/ExponentialHardeningElastoplasticMaterial.cpp
+++ b/ALE_Finite_Strain_Plasticity/src/ExponentialHardeningElastoplasticMaterial.cpp
@@ -49,17 +49,17 @@ namespace PlasticityLab {
 
   template <int dim, typename Number>
   std::vector<Number> ExponentialHardeningElastoplasticMaterial<dim, Number>::get_state_parameters(
-      const point_index_t &point_index,
-      const Tensor<2, dim, Number> &reference_transformation) const {
+      const point_index_t &,
+      const Tensor<2, dim, Number> &) const {
     throw NotImplementedException();
   }
 
   template <int dim, typename Number>
   void ExponentialHardeningElastoplasticMaterial<dim, Number>::
   set_state_parameters(
-        const point_index_t &point_index,
-        const std::vector<Number> &state_parameters,
-        const Tensor<2, dim, Number> &reference_transformation) {
+        const point_index_t &,
+        const std::vector<Number> &,
+        const Tensor<2, dim, Number> &) {
     throw NotImplementedException();
   }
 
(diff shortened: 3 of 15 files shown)


marcfehling:
Thank you for the quick response! Your changes looks reasonable.

Unfortunately, the program still triggers some assertions.
```
$ mpirun -np 2 ./PlasticityLab 

--------------------------------------------------------
An error occurred in line <2297> of file <dealii/include/deal.II/lac/trilinos_vector.h> in function
    dealii::TrilinosWrappers::MPI::Vector& dealii::TrilinosWrappers::MPI::Vector::operator=(dealii::TrilinosScalar)
The violated condition was: 
    !has_ghost_elements()
Additional information: 
    You are trying an operation on a vector that is only allowed if the
    vector has no ghost elements, but the vector you are operating on does
    have ghost elements.
    
    Specifically, there are two kinds of operations that are typically not
    allowed on vectors with ghost elements. First, vectors with ghost
    elements are read-only and cannot appear in operations that write into
    these vectors with the exception of setting the entire vector to zero.
    Second, reduction operations (such as computing the norm of a vector,
    or taking dot products between vectors) are not allowed to ensure that
    each vector element is counted only once (as opposed to once for the
    owner of the element plus once for each process on which the element
    is stored as a ghost copy).
    
    See the glossary entry on 'Ghosted vectors' for more information.

Stacktrace:
-----------
#0  ./PlasticityLab: dealii::TrilinosWrappers::MPI::Vector::operator=(double)
#1  ./PlasticityLab: PlasticityLab::PlasticityLabProg<2, double>::run()
#2  ./PlasticityLab: main
--------------------------------------------------------

```

bangerth:
Perhaps we can merge this PR by itself already given that it fixes real issues that prevent the code to run in sequential computations?

bangerth (approved):

bangerth:
For the record, the failing assertion is in a statement of the form
```
  vector = scalar;
```
which is only allowed if the scalar is zero and if `vector` does not have ghost entries. In fact, with the current deal.II version, it is also allowed if `scalar==0` and if the vector does have ghost entries. So I suspect that this happens in a statement like
```
    therm_nonlinear_system.previous_deformation = ambient_temperature;
```
in which the right hand side is nonzero. For cases like this, you need to first create a fully distributed vector, assign it the `ambient_temperature` value, and then assign that vector to the ghosted vector here.

bangerth:
Let me merge this so we make some progress towards a working code gallery!

bangerth:
#254 fixes two of the places where we modify ghosted entries. There is at least one more place, but I need to sleep so it will have to wait (help appreciated!). That place is here, PlasticityLabProgDrivers.cpp:
```
      mesh_motion_nonlinear_system.advance_time(time_increment, rho_infty, false);
      mesh_motion_nonlinear_system.previous_deformation = mesh_motion_nonlinear_system.current_increment;
      mesh_motion_nonlinear_system.previous_deformation *= -1;
```
`previous_deformation` is a ghosted vector. We used to, but no longer, allow it to be multiplied by -1.

maieneuro:
Thanks! I'll look into the ghost element issues tonight.

PR #254, opened 2026-07-03
Fix two places where we modify a ghosted vector.

bangerth:
See https://github.com/dealii/code-gallery/pull/252#issuecomment-4874561581 One could perhaps do that differently, but I think these are acceptable approached.

There is one more place (at least) to fix, see the other comments in #252, but it's too late here.

@maieneuro FYI.

### DIFF
--- a/ALE_Finite_Strain_Plasticity/src/NewtonStepSystem.h
+++ b/ALE_Finite_Strain_Plasticity/src/NewtonStepSystem.h
@@ -41,20 +41,31 @@ namespace PlasticityLab {
         dof_system.locally_owned_dofs,
         mpi_communicator);
       current_increment.reinit(
+        dof_system.locally_owned_dofs,
         dof_system.locally_relevant_dofs,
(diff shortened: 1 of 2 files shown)


ALE_Finite_Strain_Plasticity/src/NewtonStepSystem.h
```diff
@@ -41,22 +41,33 @@ namespace PlasticityLab {
...
         mpi_communicator);
 
       previous_time_derivative.reinit(
+        dof_system.locally_owned_dofs,
         dof_system.locally_relevant_dofs,
         mpi_communicator);
       previous_second_time_derivative.reinit(
+        dof_system.locally_owned_dofs,
         dof_system.locally_relevant_dofs,
         mpi_communicator);
```
bangerth, line 62: Not strictly necessary, but we have been using the two-indexset form of these functions.

maieneuro:
I added a second commit to this: #255

PR #255, opened 2026-07-04
ALE_Finite_Strain_Plasticity: fix MPI aborts from writing into ghosted vectors

maieneuro:
## Summary

Under MPI, the ALE_Finite_Strain_Plasticity example aborted during time
stepping with the deal.II assertion `!has_ghost_elements()`. The cause was
several places that wrote arithmetic results directly into
`previous_deformation`, which is a vector with ghost entries and therefore
read-only.

This PR fixes all such writes so the example builds cleanly and runs to
completion under MPI.

## Changes

- **Fix two places where we modify a ghosted vector** (6c6caa3): initial
  fix for the ghosted-vector writes surfaced by the assertion.
- **Fix remaining ghosted-vector writes** (b135642): two further writes into
  `previous_deformation` remained:
  - `mesh_motion`: `previous_deformation = current_increment; *= -1;`
  - `thermal`:     `previous_deformation += current_increment;`

  Both are now expressed as named `NewtonStepSystem` member functions
  (`add_current_increment_to_previous_deformation()` and
  `set_previous_deformation_to_negative_current_increment()`) that perform the
  arithmetic in fully-distributed temporary vectors and only assign the result
  back into the ghosted vector, which triggers the ghost-value communication.
  `advance_time()` reuses the first of these, removing a duplicated update block.

## Verification

Builds cleanly and runs to completion (all 80 time steps) under
`mpirun -np 2` with no ghost-element assertions.

### DIFF
--- a/ALE_Finite_Strain_Plasticity/src/NewtonStepSystem.h
+++ b/ALE_Finite_Strain_Plasticity/src/NewtonStepSystem.h
@@ -122,21 +122,40 @@ namespace PlasticityLab {
 
       }
       
-      // Update the deformation vector with the computed
-      // increment. Because 'previous_deformation' is a vector with
-      // ghost entries, it is a read-only vector and we cannot write
-      // into it -- so we have to create a temporary vector first and
-      // only copy it at the end:
-      {
-        _locally_owned_previous_deformation = previous_deformation;
-        _locally_owned_current_increment    = current_increment;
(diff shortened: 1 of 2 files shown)


bangerth:
Thanks for making my commit nicer by introducing the helper functions.

@marcfehling Since I wrote one of the commits, would you care to look this over and merge?

marcfehling:
Thank you all! The patch runs on my machine in parallel. I will go ahead and merge.

PR #257, opened 2026-07-29
CI: Build on noble and drop jammy. Also build master on resolute.

marcfehling:
This PR updates the github actions workflows to use noble, and we will also build the master image on resolute. This patch will drop the checks for jammy to be mindful regarding github actions resources.

I had to update libprecice as only version >=3.4.0 provides packages for resolute.

Further, I fixed a code gallery program because of https://github.com/dealii/dealii/pull/20082.

### DIFF
--- a/.github/workflows/linux.yml
+++ b/.github/workflows/linux.yml
@@ -83,9 +83,15 @@ jobs:
       - name: Test coupled_laplace_problem
         run: |
           cd coupled_laplace_problem
-          (./coupled_laplace_problem 2>&1 & ./fancy_boundary_condition >fbc.log)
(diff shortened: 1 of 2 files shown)


marcfehling:
CI on resolute failed with `sed: can't read solution-10.vtk: No such file or directory`

It looks like the script continues before `coupled_laplace_problem` has finished. Strange that it doesn't happen on noble.

https://github.com/dealii/code-gallery/blob/0a59915c26f25e5137a32376c3972d4c84fdf993/.github/workflows/linux.yml#L83-L88

marcfehling:
Seems to work now!

masterleinad (approved):

